PR #91, opened 2024-04-29
Icontract storage fix

sevenearths:


sevenearths:
> @sevenearths please update with main and adapt to the breaking changes from #92

Done!
But this needs to be looked over to see if it's changes are relevant

Issue #86, opened 2024-04-27
GenLayer Simulator Backend Fixes


cristiam86:
- An Intelligent Contract function shouldn't need a fromAddress input parameter to run
- An Intelligent Contract function shouldn't need an equivalence principle to run
- An Intelligent Contract function shouldn't need the async keyword to be defined
- The Contract state is mixing the input parameters and the self.<variable> should be updated an returned in the get_contract_state call

cristiam86:
Closed in: https://github.com/yeagerai/genlayer-simulator/pull/90

PR #93, opened 2024-04-29
86 genlayer simulator backend fixes

sevenearths:
Only one remaining is:

- The Contract state is mixing the input parameters and the self. should be updated an returned in the get_contract_state call

Please read [this comment](https://github.com/yeagerai/genlayer-simulator/issues/86#issuecomment-2083855120)


sevenearths:
> @sevenearths please update with main and adapt to the breaking changes from #92

Done!

rpc/server.py
```diff
@@ -400,11 +402,11 @@ async def call_contract_function(
...
     )
 
-    return {
-        "status": "success",
-        "message": f"Function '{function_name}' called on contract at {contract_address} with args {args}.",
-        "execution_output": execution_output,
-    }
+    return_data['message'] = f"Function '{function_name}' called on contract at {contract_address} with args {args}."
+
+    return_data['data'] = execution_output
```
cristiam86, line 407:
```suggestion
    return_data['execution_output'] = execution_output
```
just to keep the format the same for now

genvm/server.py
```diff
@@ -164,7 +161,64 @@ def get_icontract_schema(icontract: str) -> dict:
                     if hasattr(stmt.annotation, "id") and hasattr(stmt.target, "id"):
                         variables[stmt.target.id] = stmt.annotation.id
 
-    return {"class": class_name, "methods": methods, "variables": variables}
+    return {"class":class_name, "methods": methods, "variables": variables}
+
+
+@jsonrpc.method("deploy_contract")
```
cristiam86, line 167: why these methods appear as added if they are already in main just like that?

cristiam86 (changes requested):
Please check the issues mentioned

cristiam86:
Not needed anymore

PR #99, opened 2024-04-30
98 Fixes on executing contract

cristiam86:


jmlago (approved):
working on my end for wizzard

PR #123, opened 2024-05-03
38 add logger detail

bradleySuira:
#38


frontend/src/components/Simulator/NodeLogs.vue
```diff
@@ -4,13 +4,22 @@ import { webSocketClient } from '@/utils'
...
+const colorMap: Record<string, string> = {
+  'info': 'text-blue-500',
+  'error': 'text-red-500',
+  'warning': 'text-yellow-500',
+  'success': 'text-green-500'
+}
 
-
+const getDataText = (data: any) => {
+  if (typeof data !== 'string') return JSON.stringify(data)
```
cristiam86, line 15: @bradleySuira what if we do `JSON.stringify(data, null, 2)` to print it a little bit better?

genvm/server.py
```diff
@@ -7,35 +7,39 @@
...
 def leader_executes_transaction(contract_code: str, node_config: dict) -> dict:
 
     delete_recipts()
 
-    return_data = {"status": "error", "data": None}
+    msg = MessageHandler(app, socketio)
 
     icontract_file, _, _, leader_recipt_file = transaction_files()
 
+    msg.debug_response("icontract", contract_code)
```
cristiam86, line 40:
```suggestion
    msg.debug_response("Contract Code", contract_code)
```

genvm/server.py
```diff
@@ -76,11 +80,17 @@ def validator_executes_transaction(
...
 
     delete_recipts()
+    
+    msg = MessageHandler(app, socketio)
 
     return_data = {"status": "error", "data": None}
 
     icontract_file, recipt_file, _, _ = transaction_files()
 
+    msg.debug_response("icontract", icontract)
```
cristiam86, line 90:
```suggestion
    msg.debug_response("Contract Code", contract_code)
```

rpc/utils.py
```diff
@@ -0,0 +1,27 @@
+import re
+import string
+import random
+
+
+def create_new_address() -> str:
```
cristiam86, line 6: @sevenearths these two methods about the address look like they should be at the protocol level. @jmlago do you think they should be in the GenVM?

cristiam86 (changes requested):

common/config/logging.dev.json
```diff
@@ -0,0 +1,47 @@
...
+    "formatters": {
+        "default": {
+            "format": "[%(asctime)s] %(levelname)s | %(module)s >>> %(message)s",
+            "datefmt": "%B %d, %Y %H:%M:%S %Z"
+        },
+        "simple": {
+            "format": "%(levelname)s | %(message)s"
+        }
+    },
+    "handlers": {
```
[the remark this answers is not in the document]
sevenearths: Yup. Your right. I'll fix that.

frontend/src/components/Simulator/NodeLogs.vue
sevenearths: I'll add that

genvm/server.py
sevenearths: Updated!

genvm/server.py
sevenearths: There is no variable called `contract_code`.

sevenearths (changes requested):
@bradleySuira I think the frontend should poll `get_providers_and_models` every time a user wants to edit or create a node so it gets the latest information
IMG20240506111803_BURST000_COVER

The code is missing the `get_icontract_schema_for_code` function

When `deploy_intelligent_contract` is called it's missing the class method

sevenearths (approved):

cristiam86:
I fixed the get_contract_state response format to match what the frontend is expecting. @bradleySuira, please take a look since I can't ask for your review here because you are the owner of the PR.

sevenearths (approved):

sevenearths (approved):

sevenearths (approved):

sevenearths (approved):

sevenearths (approved):

cristiam86 (approved):

PR #137, opened 2024-05-09
116 simulator frontend  improve the input of constructor parameters

bradleySuira:


cristiam86 (changes requested):
It looks nice but we need some kind of message (could be the same toast) when the JSON parsing fails and the user needs to correct it.
For example, this one fails but it doesn't say anything: `{ "args": True}`
Also, we need to accept the possibility that there aren't any constructor parameters. In that case it could just be no json or an empty json (i.e. {})

bradleySuira:
@cristiam86 updated the validations abut args and empty constructor params

cristiam86 (approved):

PR #146, opened 2024-05-13
139 sim eq delete internal variables end of block

sevenearths:
code checking functionality checks for the following situations:
- Valid Python code
- Class passed with code exists in code
- Eq block is instantiated incorrectly (i.e. `a = EquivalencePrinciple()`)
- Class variables are not assigned a value in eq blocks
- eq block values are not used outside of the eq block


examples/contracts/wizzard_of_coin.py
```diff
@@ -34,10 +34,13 @@ async def ask_for_coin(self, request: str) -> None:
...
                 eq_principle="The result['give_coin'] has to be exactly the same",
             )
-            result_clean = result.replace("True", "true").replace("False", "false")
-            output = json.loads(result_clean)
 
-            if output["give_coin"] is True:
+            try:
+                result = json.loads(result)
+            except json.JSONDecodeError:
+                raise Exception("The validator did not return valid JSON", result)
```
cristiam86, line 41: This is not the validator returning a valid JSON, is the LLM

examples/contracts/wizzard_of_coin.py
```diff
@@ -1,5 +1,5 @@
 import json
-from genvm.base.equivalence_principle import call_llm_with_principle
+from genvm.base.equivalence_principle import EquivalencePrinciple, call_llm_with_principle
 
 
 class WizzardOfCoin:
```
cristiam86, line 5: Rename to Wizard

scripts/debug_simulator.py
```diff
@@ -69,15 +69,15 @@
 if 'result' in fund_account_result and 'address' in fund_account_result['result']['data']:
     print('Account funded! ('+str(balance)+')')
 
-
 # Your hardcoded values
 contract_file_path = 'examples/contracts/wizzard_of_coin.py'
 function_to_execute = 'WizzardOfCoin.ask_for_coin'
+class_name = 'WizzardOfCoin'
```
cristiam86, line 75: Rename to Wizard

genvm/code_enforcement.py
```diff
@@ -0,0 +1,165 @@
...
+
+class ClassExistsVisitor(ast.NodeVisitor):
+    def __init__(self, class_name):
+        self.class_name = class_name
+        self.class_exists = False
+
+    def visit_ClassDef(self, node):
+        if node.name == self.class_name:
+            self.class_exists = True
+        self.generic_visit(node)
```
cristiam86, line 55: @sevenearths please add some comments or explain how this approach works.

cristiam86 (changes requested):

genvm/code_enforcement.py
sevenearths: The file has been updated accordingly => https://github.com/yeagerai/genlayer-simulator/pull/146/commits/529a9b991a0f8be0df5b1bc9ba50b55832cd5efe

jmlago (commented):
LGTM

I really like the approach, good job! 

Also note that this is dealing with syntax which is an ongoing task, here and everywhere else... So what will happen is that different edgecases will lead to modification of the filters / addition of new filters + tests.

cristiam86 (approved):
Tested with all the examples and it works

sevenearths:
> LGTM
> 
> I really like the approach, good job!
> 
> Also note that this is dealing with syntax which is an ongoing task, here and everywhere else... So what will happen is that different edgecases will lead to modification of the filters / addition of new filters + tests.

Thanks :)

PR #158, opened 2024-05-14
156-simulator---allow-to-specify-validators-provide-on-create-random-validators

cristiam86:
Added "exclusive providers" params to create random validators based only on that list.
The CLI requires that the validators be created only based on user selection. One option could have been to remove the ollama's models if the user doesn't select them, but that seems like it could impact other users' applications.

### DIFF
--- a/consensus/nodes/create_nodes.py
+++ b/consensus/nodes/create_nodes.py
@@ -17,9 +17,9 @@ def get_random_provider_using_weights(defaults):
     # remove providers if no api key
     provider_weights = defaults['provider_weights']
     default_value = '<add_your_api_key_here>'
-    if ('openai' in provider_weights and provider_weights['openai'] > 0 and ('OPENAIKEY' not in os.environ or os.environ['OPENAIKEY'] == default_value)):
+    if ('openai' in provider_weights and ('OPENAIKEY' not in os.environ or os.environ['OPENAIKEY'] == default_value)):
         provider_weights.pop('openai')
-    if ('heuristai' in provider_weights and provider_weights['heuristai'] > 0 and('HEURISTAIAPIKEY' not in os.environ or os.environ['HEURISTAIAPIKEY'] == default_value)):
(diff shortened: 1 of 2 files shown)


consensus/nodes/create_nodes.py
```diff
@@ -17,9 +17,9 @@ def get_random_provider_using_weights(defaults):
     # remove providers if no api key
     provider_weights = defaults['provider_weights']
     default_value = '<add_your_api_key_here>'
-    if 'OPENAIKEY' not in os.environ or os.environ['OPENAIKEY'] == default_value:
+    if ('openai' in provider_weights and provider_weights['openai'] > 0 and ('OPENAIKEY' not in os.environ or os.environ['OPENAIKEY'] == default_value)):
```
sevenearths, line 20: `provider_weights['openai'] > 0` I think we are making the if; overly complex and less legible. Let the random weights function handle this.

consensus/nodes/create_nodes.py
```diff
@@ -17,9 +17,9 @@ def get_random_provider_using_weights(defaults):
...
         provider_weights.pop('openai')
-    if 'HEURISTAIAPIKEY' not in os.environ or os.environ['HEURISTAIAPIKEY'] == default_value:
+    if ('heuristai' in provider_weights and provider_weights['heuristai'] > 0 and('HEURISTAIAPIKEY' not in os.environ or os.environ['HEURISTAIAPIKEY'] == default_value)):
```
sevenearths, line 22: `provider_weights['heuristai'] > 0` I think we are making the if; overly complex and less legible. Let the random weights function handle this.

consensus/nodes/create_nodes.py
cristiam86: @sevenearths  Do you suggest to assign 0 to it?

sevenearths (approved):

PR #164, opened 2024-05-15
fix: added spinner for loading constructor params and handle errors w…

bradleySuira:
…hen try to load the contract schema from code


cristiam86 (commented):
@bradleySuira The loading is cool but do you think that it would be possible to remove the loading when we get te socket message about the transaction being completed? 
Also, we need the loading on the execute transactions as well

cristiam86 (approved):

PR #167, opened 2024-05-16
160 simulator   UI fixes

bradleySuira:


frontend/src/components/JsonViewer/json-box.vue
```diff
@@ -0,0 +1,160 @@
...
+      dataType = JsonRegexp;
+    }
+    const complex =
+      this.keyName &&
+      this.value &&
+      (Array.isArray(this.value) ||
+        (typeof this.value === "object" &&
+          Object.prototype.toString.call(this.value) !== "[object Date]"));
+
+    if (!this.previewMode && complex) {
```
cristiam86, line 86: @bradleySuira could you please add some comments about what is this doing and why is it needed?

frontend/src/components/JsonViewer/json-box.vue
```diff
@@ -0,0 +1,160 @@
...
+        evt.initEvent("resized", true, false);
+        this.$el.dispatchEvent(evt);
+      }
+    },
+  },
+  render() {
+    let elements = [];
+    let dataType;
+
+    if (this.value === null || this.value === undefined) {
```
cristiam86, line 57: @bradleySuira this can be improved by having a mapping object with keys equals to the this.value and values equals to the dataTypes.

frontend/src/index.d.ts
```diff
@@ -1,2 +1,20 @@
 // declare module '@globalhive/vuejs-tour'
-declare module 'vue3-json-viewer';
\ No newline at end of file
+declare module 'vue3-json-viewer' {
+    import { AllowedComponentProps, App, Component, ComponentCustomProps, VNodeProps } from 'vue'
+    interface JsonViewerProps {
+        value: Object | Array<any> | string | number | boolean; //对象
+        expanded: boolean; //是否自动展开
+        expandDepth: number; //展开层级
+        copyable: boolean | object; //是否可复制
```
cristiam86, line 8: @bradleySuira what are this comments?? :D

cristiam86 (changes requested):

cristiam86:
@bradleySuira the debounce doesn't seem to work. With every key press in the code, it loads the constructor parameters.

frontend/src/index.d.ts
bradleySuira: fixed, we ended using a portion of a package/component because was not able to extend and don't want to create our own from scratch

frontend/src/components/Simulator/CodeEditor.vue
```diff
@@ -74,4 +74,8 @@ watch(
   </div>
 </template>
 
-<style scoped></style>
+<style>
```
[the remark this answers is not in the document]
bradleySuira: will be for now, I found the element to remove the scroll bot breaks a lot of things to will be empty style

frontend/src/components/JsonViewer/json-box.vue
bradleySuira: added

frontend/src/components/JsonViewer/json-box.vue
bradleySuira: tried but in this case we are not using always `typeof` and we have some other cases for complex types like `Array` and `Dates`

bradleySuira:
changed debounce and tested locally @cristiam86

cristiam86:
> changed debounce and tested locally @cristiam86

I see the debounce working because the API call is made only after 3 seconds. The problem is that the loading in the constructor parameters triggers on every key press without debouncing.

cristiam86 (changes requested):
Please add the debounce at the constructor parameters level so the loading is not shown unless the backend call is made.
Also, please check if the keyName and previewMode parts in the json-box components are needed: https://github.com/yeagerai/genlayer-simulator/pull/167/files#diff-b7461a1c9d7ebf3f6f680ce66716d289008d546a50617877e4c8f147fa9b7087R81-R141

cristiam86 (approved):

PR #174, opened 2024-05-21
150 sim add transactions audit table

sevenearths:


database/tests/test_transactions.py
```diff
@@ -0,0 +1,82 @@
...
+    except psycopg2.errors.CheckViolation:
+        assert True
+
+
+# TODO: This
+def test_get_transaction_for_address_that_doesnt_exist():
+    pass
+
+
+def test_get_transaction_for_transacttion_that_does_not_exist():
```
cristiam86, line 50:
```suggestion
def test_get_transaction_for_transaction_that_does_not_exist():
```

rpc/server.py
```diff
@@ -114,38 +102,62 @@ def create_account() -> dict:
...
         )
+        connection.commit()
+        cursor.close()
+        connection.close()
     except Exception as e:
         return msg.error_response(exception=e)
+
+    connection = get_genlayer_db_connection()
+    cursor = connection.cursor()
+    cursor.execute("SELECT * FROM current_state;")
```
cristiam86, line 113: @sevenearths what are we doing with this query?

cristiam86 (commented):
I have a few comments:
1. How should tests run? Please include some guidance in the Readme.
2. It doesn't seem that we are using any testing framework.
3. These tests aren't unit tests, they are integration tests very tied to the database. I think that we should focus mainly on unit testing and a few integration tests.

what do you think @jmlago?

rpc/server.py
```diff
@@ -142,13 +142,24 @@ def fund_account(account: string, balance: float) -> dict:
...
-            "INSERT INTO transactions (from_address, to_address, data, value, type) VALUES (NULL, %s, %s, %s, 0);",
+            "INSERT INTO transactions (from_address, to_address, data, value, type) VALUES (NULL, %s, %s, %s, 0) RETURNING id;",
             (
                 current_account,
                 json.dumps({"action": "create_account", "initial_balance": balance}),
                 balance,
             ),
         )
+        new_transaction_id = cursor.fetchone()
+        cursor.execute("SELECT * FROM transactions WHERE id = %s;", (new_transaction_id,))
```
cristiam86, line 153: @sevenearths do we need to fetch the transaction again? couldn't we trust that if we get the ID, then the transaction has been stored?

rpc/utils.py
```diff
@@ -24,4 +29,15 @@ def response_format(status:str, message:str='', data={}) -> dict:
...
+    }
+
+def db_query_data_to_json(db_cursor:psycopg2.extensions.cursor, db_query_result:dict) -> dict:
+    colnames = [desc[0] for desc in db_cursor.description]
+    dict_with_names = dict(zip(colnames, db_query_result))
+    for k, v in dict_with_names.items():
+        if isinstance(v, Decimal):
+            dict_with_names[k] = float(v)
+    for k, v in dict_with_names.items():
+        if isinstance(v, datetime.datetime):
```
cristiam86, line 41: @sevenearths can't we do both parsings in the same for?

sevenearths:
> I have a few comments:
> 
> 1. How should tests run? Please include some guidance in the Readme.
> 2. It doesn't seem that we are using any testing framework.
> 3. These tests aren't unit tests, they are integration tests very tied to the database. I think that we should focus mainly on unit testing and a few integration tests.
> 
> what do you think @jmlago?

The tests are now in separate branch.

cristiam86 (approved):

PR #176, opened 2024-05-22
moved the try execpt up to the clear_account_and_transactions_tables …

sevenearths:
…function in rpc/server.py

### DIFF
--- a/rpc/server.py
+++ b/rpc/server.py
@@ -94,13 +94,8 @@ def create_tables() -> dict:
 
 @jsonrpc.method("clear_account_and_transactions_tables")
 def clear_account_and_transactions_tables() -> dict:
-    msg = MessageHandler(app, socketio)
-    result = None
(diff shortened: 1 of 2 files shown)


rpc/server.py
```diff
@@ -94,9 +94,13 @@ def create_tables() -> dict:
 
 @jsonrpc.method("clear_account_and_transactions_tables")
 def clear_account_and_transactions_tables() -> dict:
-    result = clear_db_tables(["current_state", "transactions"])
-    app.logger.info(result)
-    return {"status": result}
+    msg = MessageHandler(app, socketio)
+    result = None
+    try:
```
jmlago, line 99: I would remove this try and then make that `clear_db_tables` behaves correctly, so first check if there are tables, if that is the case then clear, else do nothing.

cristiam86 (approved):

PR #183, opened 2024-05-24
added an action that only runs when a pr is opened, synchronize, reop…

sevenearths:
…ened


.github/workflows/pytest_unit_pr_approval.yml
```diff
@@ -0,0 +1,39 @@
+name: PyTest - Unit - PR approval
+
+on:
+  pull_request:
+    branches:
```
cristiam86, line 5: @sevenearths this is not the right config for PR approval

cristiam86 (changes requested):

.github/workflows/pytest_unit_pr_approval.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.12.2
+
+      - name: Install dependencies
```
cristiam86, line 22: @sevenearths we should run all the tests with one command. Also, the tests should be located in the root folder of the repo. If we have something like `tests/test_rpc/` then we should do something like `cd tests && pytest .`

cristiam86 (changes requested):

.github/workflows/pytest_unit_pr_approval.yml
sevenearths: It might be best to wait for the refactor first

PR #187, opened 2024-05-24
added and e2e test for wizard of coin

sevenearths:


.github/workflows/pytest_e2e_pr_approval.yml
```diff
@@ -0,0 +1,63 @@
+name: CI
+
+on:
```
cristiam86, line 3: @sevenearths to run actions on PR approvals, the code should be:
on:
  pull_request_review:
    types: [submitted]

.github/workflows/pytest_e2e_pr_approval.yml
```diff
@@ -0,0 +1,63 @@
...
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip install -r requirements.test.txt
+
+      - name: Run tests
+        run: pytest tests/e2e/
+
+      - name: Shutdown Docker Compose
+        if: always()
+        run: docker-compose down
```
cristiam86, line 63: @sevenearths have you tested this command? shouldn't be `docker compose ...`?

rpc/server.py
```diff
@@ -620,6 +620,11 @@ def ping() -> dict:
     return {"status": "OK"}
 
 
+@app.route('/github-actions/ping')
```
cristiam86, line 623: @sevenearths why don't you use the existing ping endpoint?

rpc/tests/test_validator_endpoints.py
```diff
@@ -1,4 +1,4 @@
-from rpc.tests.base import payload, post_request
+from common.testing.requests import payload, post_request
```
cristiam86, line 1: @sevenearths aren't these e2e endpoints? they should be with the rest of the e2e tests, right?

cristiam86 (changes requested):

.github/workflows/pytest_e2e_pr_approval.yml
sevenearths: Your right. I'll fix that

.github/workflows/pytest_e2e_pr_approval.yml
sevenearths: Again. Your right **(old habits die hard lol)**. I'll fix that.

rpc/server.py
sevenearths: There was a reason, but I can't remember what it it. Have switched it over to using the ping end point

rpc/tests/test_validator_endpoints.py
sevenearths: Right again. I'll move them

rpc/server.py
sevenearths: I'll do it as a jsonrpc call. Np

.github/workflows/pytest_e2e_pr_approval.yml
```diff
@@ -0,0 +1,77 @@
+name: CI
+
+on:
+  pull_request:
+    branches:
+      - staging
+    types:
+      - opened
+      - synchronize
```
cristiam86, line 9: @sevenearths lets avoid running the tests on every PR update. I think we should do it when it is opened and approved

cristiam86 (changes requested):

.github/workflows/pytest_e2e_pr_approval.yml
sevenearths: Those are there to re-run it when the code has been updated on a pr

cristiam86: No, we said that the tests are not going to run the tests on every PR updated, only when opened and approved

sevenearths:
> Good job @sevenearths as a final step, could you please move the common/testing folder inside the tests folders to be like /tests/common/response.py

Sure. Not a problem

PR #194, opened 2024-05-30
193 sim   create tables in docker compose init

cristiam86:
A command for the Postgres docker-compose service has been added to:
1. Wait for the database to be ready
2. Run the needed commands to create the tables if they don't exist

There are also some files that have been relocated to the database root folder.

Lastly, some format has been added when modifying the existing file rpc/server.py. Line 14 is the only actual change.


jmlago (approved):
working on my end

PR #199, opened 2024-06-03
fix: fixed execute transactions abi update and selected method

bradleySuira:


cristiam86:
The "debounce" for the constructor parameters makes it weird when switching tabs. Everything else changes fast but not the constructor parameters.
I would say that we need to avoid the debounce on this occasion.

bradleySuira:
> The "debounce" for the constructor parameters makes it weird when switching tabs. Everything else changes fast but not the constructor parameters. I would say that we need to avoid the debounce on this occasion.

fixed now @cristiam86 the debounce is applied only when contract content change

cristiam86:
> > The "debounce" for the constructor parameters makes it weird when switching tabs. Everything else changes fast but not the constructor parameters. I would say that we need to avoid the debounce on this occasion.
> 
> fixed now @cristiam86 the debounce is applied only when contract content change

Now I don't see the loading when changing contract tabs but after 3 seconds, the call is made anyway and the constructor parameters section blinks. Do you think there is something to do?

cristiam86 (changes requested):
There is another issue when refreshing from scratch. The first Contract that you choose doesn't show the Constructor Parameters

https://github.com/yeagerai/genlayer-simulator/assets/4550355/ddb72923-94fb-4f33-99e0-101c891a9844

cristiam86 (approved):

PR #201, opened 2024-06-04
200 SIM-FE-Allow the user to reset all their contracts and local storage

bradleySuira:
- Added new option in settings screen to reset the contracts storage


frontend/src/views/Simulator/SettingsView.vue
```diff
@@ -422,5 +468,23 @@ const handleCreateNewValidator = async () => {
         </button>
       </div>
     </Modal>
+    <Modal :open="resetStorageModalOpen" @close="closeResetStorageModal">
+      <div class="flex flex-col">
+        <div class="flex justify-between">
+          <div class="text-xl">Reset Simulator Storage</div>
+        </div>
+        <div class="flex justify-between font-bold bg-slate-100 p-2 mt-4">
+          Are you sure you want to reset the simulator storage?
```
cristiam86, line 477: @bradleySuira please add more information about what this will do to all your contracts and stuff.
As far as I understand, it will remove all the users contracts as well right? Is there a way to avoid it? by setting some parameter in the database, for example?

cristiam86 (changes requested):
From now on, the PRs should point to staging and please make the title the same as the issue title.

frontend/src/views/Simulator/SettingsView.vue
bradleySuira: fixed as requested

frontend/src/utils/store.ts
```diff
@@ -2,13 +2,21 @@ import { useMainStore } from '@/stores'
...
   const mainStore = useMainStore()
-  // await db.contractFiles.clear()
-  // localStorage.setItem('mainStore.contractsModified', '')
+  const contracts = await db.contractFiles.toArray()
   if (
-    (await db.contractFiles.count()) === 0 &&
-    !localStorage.getItem('mainStore.contractsModified')
+    contracts
+      .filter((c) => (c.example && !c.updatedAt) || examplesNames.includes(c.name))
+      .map((c) => c.id).length === 0
```
cristiam86, line 19: @bradleySuira what is this map doing? do we need it?

cristiam86: Also, the comparison with the name doesn't seem reliable for this decision

frontend/src/utils/store.ts
```diff
@@ -29,18 +38,19 @@ export const setupStores = async () => {
   }
 
   mainStore.deployedContracts = await db.deployedContracts.toArray()
-  if ( mainStore.accounts.length < 1) {
+  if (mainStore.accounts.length < 1) {
     await mainStore.generateNewAccount()
   } else {
-    mainStore.accounts = localStorage.getItem('mainStore.accounts') ?  (localStorage.getItem('mainStore.accounts') || '').split(',') : []
+    mainStore.accounts = localStorage.getItem('mainStore.accounts')
+      ? (localStorage.getItem('mainStore.accounts') || '').split(',')
```
cristiam86, line 45: @bradleySuira should we access local storage from here? Aren't we using the state management for the accounts as well?

frontend/src/views/Simulator/SettingsView.vue
```diff
@@ -408,5 +452,29 @@ const handleCreateNewValidator = async () => {
...
+        </div>
+        <div class="flex flex-col p-2 mt-2">
+
+          <div class="py-2 w-full">
+            All the contracts examples that are not modified will be set to their initial state.
```
cristiam86, line 466: @bradleySuira, as we discussed, the modified examples are going to be also reset so we need to specify that as well

cristiam86 (changes requested):

frontend/src/utils/store.ts
bradleySuira: In this case we feed the accounts state form local storage, that's only to not create a lot of new accounts every time
- First load we don't have any account so we call to the PRC to create a new on
- Then we load the existing accounts from local storage and the user can use them or add new accounts

the only issue is that they are not synced with the Database if the user drop the storage or reset the containers but for now I think is ok not sure in the future because accounts should be created in the client side based on seed phrase maybe if we will use EOA

frontend/src/utils/store.ts
bradleySuira: @cristiam86 
- removed map we don't need it
- we need to keep the names just in case for old browser storage because now we are adding the example field inside the db
-

frontend/src/stores/main.ts
```diff
@@ -96,6 +95,31 @@ export const useMainStore = defineStore('mainStore', {
         console.error
         return null
       }
+    },
+    async resetStorage(): Promise<void> {
+      try {
+        const contracts = await db.contractFiles.toArray()
+        const idsToDelete = contracts.filter((c) => c.example && !c.updatedAt).map((c) => c.id)
```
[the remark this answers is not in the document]
bradleySuira: fixed

frontend/src/views/Simulator/SettingsView.vue
bradleySuira: fixed

frontend/src/views/Simulator/SettingsView.vue
```diff
@@ -408,5 +452,29 @@ const handleCreateNewValidator = async () => {
...
+        <div class="flex justify-between">
+          <div class="text-xl">Reset Simulator Storage</div>
+        </div>
+        <div class="flex justify-between font-bold bg-slate-100 p-2 mt-4">
+          Are you sure you want to reset the simulator storage?
+        </div>
+        <div class="flex flex-col p-2 mt-2">
+
+          <div class="py-2 w-full">
+            All the contracts examples and created contracts not updated will be set to their initial state.
```
cristiam86, line 466:
```suggestion
            All the examples of contracts and created contracts by you that have not been updated will be set to their initial state.
```

cristiam86 (changes requested):

frontend/src/stores/main.ts
```diff
@@ -96,6 +95,31 @@ export const useMainStore = defineStore('mainStore', {
         console.error
         return null
       }
+    },
+    async resetStorage(): Promise<void> {
+      try {
+        const contracts = await db.contractFiles.toArray()
+        const idsToDelete = contracts.filter((c) => !c.example && !c.updatedAt).map((c) => c.id)
```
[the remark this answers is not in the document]
bradleySuira: got it updated now

frontend/src/utils/store.ts
bradleySuira: sure in the other pr we have an `accountsStore` that has the list of accounts and the current selected account

cristiam86 (approved):

cristiam86 (changes requested):
1. Examples are not being restored
2. If there are no contracts to be reset/removed we should hide the message "All the examples will be restored..." and the note "If you want to preserve"

bradleySuira:
> 1. Examples are not being restored
> 2. If there are no contracts to be reset/removed we should hide the message "All the examples will be restored..." and the note "If you want to preserve"

Fixed now @cristiam86

fakela (approved):

tests/test_rpc/test_address_utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+def test_create_new_address_format():
+    address = create_new_address()
+    assert address.startswith("0x")
+    assert address_is_in_correct_format(address)
+
+
+def test_address_is_in_correct_format_valid():
+    valid_address = "0x" + "a" * 40
+    assert address_is_in_correct_format(valid_address) == True
```
AgustinRamiroDiaz, line 19:
```suggestion
    assert address_is_in_correct_format(valid_address)
```

tests/test_rpc/test_address_utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+def test_address_is_in_correct_format_invalid():
+    invalid_addresses = [
+        "0x" + "z" * 40,  # Invalid character 'z'
+        "0x" + "a" * 39,  # Incorrect length
+        "1x" + "a" * 40,  # Incorrect prefix
+        "0x" + "a" * 41,  # Incorrect length
+    ]
+    for address in invalid_addresses:
+        assert address_is_in_correct_format(address) == False
```
AgustinRamiroDiaz, line 30:
```suggestion
        assert not address_is_in_correct_format(address)
```

AgustinRamiroDiaz (approved):
I'm not a frontend dev so my approval is based on manual testing :+1:

fakela:
> I'm not a frontend dev so my approval is based on manual testing 👍

Same here :)

AgustinRamiroDiaz:
Merging so we can release

PR #205, opened 2024-06-04
204 SIM - FE - Check autoscrolling feature on logs

bradleySuira:


cristiam86 (changes requested):
The functionality works but please make the PR pointing to staging and set the title as the issue title

cristiam86 (changes requested):
@bradleySuira the title is still wrong

PR #216, opened 2024-06-10
206 SIM-FE-Ensure code editor always resize to max possible espace

bradleySuira:
This PR includes improvements in the code editor to handle resize of the screen and panels, to test it:
- show dev console options and resize the windows and screen
- change the size vertical and horizontal of the panels

The code editor should take all the space available


cristiam86 (changes requested):
@bradleySuira it doesn't seem like the best approach to react to a boolean variable being changed. If you resize the code editor more than once, you will set true to the variable shouldResize twice but with only one render (the first one since the second never fires a re-render because it is the same value).
The right approach should be to listen to the events being emitted to execute the resize every time

bradleySuira:
> @bradleySuira it doesn't seem like the best approach to react to a boolean variable being changed. If you resize the code editor more than once, you will set true to the variable shouldResize twice but with only one render (the first one since the second never fires a re-render because it is the same value). The right approach should be to listen to the events being emitted to execute the resize every time

Thanks @cristiam86 I pushed an update using an event listener and resize observer, we are listening for windows resize and for the container element of the editor and assign the size to it.

cristiam86 (approved):

PR #219, opened 2024-06-11
213 sim dev and prod docker config

sevenearths:
- Added `docker-compose.yml` to the gitignore so it doesn't accidentally get committed by dev commits *(i.e. `cp docker-compose.dev.yml docker-compose.yml` when working on the code)*
- Updated `contributing.md` with vscode debugging and `docker-compose.dev.yml` instructions
- Cleaned up `docker-compose.yml` removing all the things that are only needed in dev
- Cleaned up the jsonrpc and genvm Dockerfile files to use the respective scripts for hot reloading/debugging as tthis should be switchable at the container level, NOT the image level *(i.e. You shouldn't have to re-build and image if you want to use the debugger in VSCode)*


sevenearths:
@cristiam86 @jmalgo 
The other way to go with this is to have a `docker-compose.prod.yml` and a `docker-compose.dev.yml`, having `docker-compose.yml` in the gitignore and then just cp'ing either of them to `docker-compose.yml`.
Tell me your thoughts.

cristiam86 (changes requested):
@sevenearths We need to find a different approach since docker compose commands has proven to fail on different environments

sevenearths:
> @sevenearths We need to find a different approach since docker compose commands has proven to fail on different environments

Have those build issues been with JM, Bradley or you? (As everyone else would be pulling from the registry in the future)
Where are the issues? If it's to do with the bash scripts that can be rolled back to the original dockerfiles.

sevenearths:
> > > @sevenearths We need to find a different approach since docker compose commands has proven to fail on different environments
> > 
> > 
> > Have those build issues been with JM, Bradley or you? (As everyone else would be pulling from the registry in the future) Where are the issues? If it's to do with the bash scripts that can be rolled back to the original dockerfiles.
> 
> Yes, JM and Edgars had those issues.

Ah I see what you mean. I remember what happened now. If we loose the commands in the docker compose file we'll also loose the ability to quickly switch from live reload and debugging in VSCode but I suppose thats just the way it has to be.

sevenearths:
I'll put the files back to the way they were.

AgustinRamiroDiaz:
> I tried using them inside the image build but the containers (once deleted) were not switching over to live reload so they might need to be deleted in the future as the only way to do debugging is with an image rebuild unfortunately :(

@sevenearths have you tried using https://docs.docker.com/compose/file-watch/#example ? It's a relatively new feature of docker-compose which enables rebuilding containers on file changes

sevenearths:
> > I tried using them inside the image build but the containers (once deleted) were not switching over to live reload so they might need to be deleted in the future as the only way to do debugging is with an image rebuild unfortunately :(
> 
> @sevenearths have you tried using https://docs.docker.com/compose/file-watch/#example ? It's a relatively new feature of docker-compose which enables rebuilding containers on file changes

That's a good idea. I didn't know about that feature.

cristiam86:
@AgustinRamiroDiaz @YeagerAI-Bulat could you guys please look into this PR and decide whether: 
- It still makes sense and we should merge
- It still makes sense and we should merge after some adjustments
- It doesn't make sense, lets close it

YeagerAI-Bulat:
I will vote for keeping 2 separate Docker Compose files.
For small projects like this one, separate Docker Compose files or profiles can be sufficient.
But for larger projects with more complex requirements, and lots of containers using profiles within a single Docker Compose file or Makefiles might be more appropriate.

YeagerAI-Bulat:
And for the **GenLayer CLI** which docker-compose file should we use when running `init` and `up` commands? Or should it be specified?

AgustinRamiroDiaz:
> And for the **GenLayer CLI** which docker-compose file should we use when running `init` and `up` commands? Or should it be specified?

I think the production one, given that the CLI is expected to be used by end users who won't develop on the code

cristiam86:
closing this as it is not relevant anymore

PR #234, opened 2024-06-26
232 SIM Database Schema Migration

AgustinRamiroDiaz:
# What

- Implement database schema migration tooling in order to have backwards compatible releases
  - Particularly, we are using Alembic and SQLAlchemy 
- This is the first attempt to use ADRs

# Reviewer tips

- Read the ADR with a markdown renderer
- Experiment with the tests below
- Review carefully the module organization to see if there are improvements to be made

# Testing done

## From main to refactor branch

```
git checkout main
docker compose down
docker compose up --build postgres
git checkout 232-sim-database-schema-migration
docker compose up --build database-migration
```

## Starting from refactor branch

```
git checkout 232-sim-database-schema-migration
docker compose down
docker compose up --build database-migration
```

## Notes

`docker compose down` is used to delete the data from the postgres container

# Possible future improvements

- Now that we have the db model in Python written with SQLAlchemy, we can use these datatypes instead of using dictionaries to improve the safety of our code
- Now it's also clearer to see that we are missing some things in the data model
  - some columns shouldn't be nullable
  - some columns should be FK references

To keep PR size at its minimum, I've left this things out. I'll create some issues for them


decisions/README.md
```diff
@@ -0,0 +1,7 @@
+# Decisions documentation
+
+Greatly inspired by ADRs, this directory contains important decisions about the `simulator` code.
+
+# Useful documentation
+
+- https://github.com/joelparkerhenderson/architecture-decision-record
```
cristiam86, line 7: @AgustinRamiroDiaz, please, let's add the index of all the ADRs here.

decisions/handle-database-migrations.md
```diff
@@ -0,0 +1,56 @@
+# Handle database migrations with ...
```
cristiam86, line 1: @AgustinRamiroDiaz, what do you think about having a number for every ADR so we can know the order in which they were executed? In this case, it would be something like: `decisions/handle-database-migrations.md`

cristiam86 (changes requested):
Good job! only small changes

decisions/handle-database-migrations.md
AgustinRamiroDiaz: I think you meant `decisions/001-handle-database-migrations.md`
sounds good!

decisions/README.md
AgustinRamiroDiaz: Added!

backend/database_handler/models.py
```diff
@@ -0,0 +1,84 @@
...
+    __table_args__ = (PrimaryKeyConstraint("id", name="current_state_pkey"),)
+
+    id: Mapped[str] = mapped_column(String(255), primary_key=True)
+    data: Mapped[dict] = mapped_column(JSONB)
+    updated_at: Mapped[Optional[datetime.datetime]] = mapped_column(
+        DateTime(True), server_default=text("CURRENT_TIMESTAMP")
+    )
+
+
+class Transactions(Base):
```
cristiam86, line 35: @AgustinRamiroDiaz how is the status going to be created?

cristiam86 (changes requested):
Please add the steps needed to test this PR :)

AgustinRamiroDiaz:
> Please add the steps needed to test this PR :)

Sure! I've put it as a draft while its a WIP, you can unsubscribe from the notifications until I reopen it :)

backend/database_handler/models.py
AgustinRamiroDiaz: I'm planning to add the `status` field here, and then generate the automatic migration files. That would be the new way of handling schema changes
WDYT?

cristiam86: Great, is that going to be added in this PR?

AgustinRamiroDiaz: Yes! I'm working on it

cristiam86 (approved):

PR #235, opened 2024-06-27
207 SIM-e2e tests for validators configuration

bradleySuira:


frontend/src/views/Simulator/SettingsView.vue
```diff
@@ -65,7 +65,8 @@ async function handleDeleteValidator() {
     notify({
       title: 'OK',
       text: 'Validator deleted successfully',
-      type: 'success'
+      type: 'success',
+      duration: 53000
```
cristiam86, line 69: @bradleySuira is this on milliseconds? why 53000?

frontend/test/pages/SettingsPage.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  async openNewValidatorModal() {
+    const locator = By.xpath(`//button[contains(text(), 'New Validator')]`)
+    return this.driver.wait(until.elementLocated(locator), 5000).click()
+  }
+  async getValidatorsElements(): Promise<WebElement[]> {
+    return this.driver.findElements(By.xpath("//div[@data-testid = 'validator-item-container']"))
+  }
+
+  async createValidator() {
```
cristiam86, line 16: @bradleySuira this method should receive the provider, model, and stake so it is more reliable when testing the update validator

frontend/vite.config.ts.timestamp-1719509065635-6c2a7c57083b5.mjs
```diff
@@ -0,0 +1,36 @@
+// vite.config.ts
```
cristiam86, line 1: @bradleySuira should this file be commited?

cristiam86 (changes requested):

frontend/src/views/Simulator/SettingsView.vue
bradleySuira: fixed, removed the duration

frontend/vite.config.ts.timestamp-1719509065635-6c2a7c57083b5.mjs
bradleySuira: removed and added to ignore files

frontend/test/pages/SettingsPage.ts
bradleySuira: fixed now 
we have all params in the util method
```
 async createValidator({
    provider,
    model,
    stake
  }: {
    provider: string
    model: string
    stake: number
  })
```

cristiam86 (approved):

PR #238, opened 2024-06-28
Optimized docker image sizes and run them as non-root user

YeagerAI-Bulat:


docker/Dockerfile.frontend
```diff
@@ -1,16 +1,21 @@
...
 WORKDIR /app
-
-COPY /frontend/package*.json .
-
-RUN npm i
-
+COPY ./frontend/package*.json .
+RUN npm ci --cache /tmp/empty-cache && \
+    npm cache clean --force && \
+    rm -rf /tmp/empty-cache
```
AgustinRamiroDiaz, line 7: Nit: I think we can use buildkit here to mount the cache in order to have faster builds

docker/Dockerfile.genvm
```diff
@@ -1,29 +1,26 @@
...
-
-CMD python -m debugpy --listen 0.0.0.0:${GENVMDEBUGPORT} -m flask run -h 0.0.0.0 -p ${FLASK_SERVER_PORT}
+CMD ["python", "-m", "debugpy", "--listen", "0.0.0.0:${GENVMDEBUGPORT}", "-m", "flask", "run", "-h", "0.0.0.0", "-p", "${FLASK_SERVER_PORT}"]
 
 ###########START NEW IMAGE: PRODUCTION ###################
 FROM base as prod
-
-CMD flask run -h 0.0.0.0 -p ${FLASK_SERVER_PORT}
\ No newline at end of file
+CMD ["flask", "run", "-h", "0.0.0.0", "-p", "${FLASK_SERVER_PORT}"]
```
AgustinRamiroDiaz, line 26: @YeagerAI-Bulat have you checked that this continues to work as expected? I remember there are some differences between the shell and exec forms when handling envvars

https://docs.docker.com/reference/dockerfile/#variable-substitution

docker/Dockerfile.rpc
```diff
@@ -1,26 +1,29 @@
...
-CMD python -m debugpy --listen 0.0.0.0:${RPCDEBUGPORT} -m flask run -h 0.0.0.0 -p ${FLASK_SERVER_PORT}
+CMD ["python", "-m", "debugpy", "--listen", "0.0.0.0:${RPCDEBUGPORT}", "-m", "flask", "run", "-h", "0.0.0.0", "-p", "${FLASK_SERVER_PORT}"]
 
-###########START NEW IMAGE: PRODUCTION ###################
+# Start the new image for production
 FROM base as prod
 
-CMD flask run -h 0.0.0.0 -p ${FLASK_SERVER_PORT}
\ No newline at end of file
+CMD ["flask", "run", "-h", "0.0.0.0", "-p", "${FLASK_SERVER_PORT}"]
```
AgustinRamiroDiaz, line 29: Same as previous comment

AgustinRamiroDiaz (commented):
Changes look good

@YeagerAI-Bulat Do you think adding users with less scope is worth the complexity? Could you explain the benefits of reducing the permission scope of the users?
I understand that it's better in terms of having the minimum scope possible, but given that the containers are isolated I don't see much benefit

YeagerAI-Bulat:
> Changes look good
> 
> @YeagerAI-Bulat Do you think adding users with less scope is worth the complexity? Could you explain the benefits of reducing the permission scope of the users? I understand that it's better in terms of having the minimum scope possible, but given that the containers are isolated I don't see much benefit
Adding users with less scope can indeed be complex, but it enhances security by minimizing potential attack surfaces. Even though containers are isolated, a compromised container could still be exploited to affect the host system. Reducing permission scope limits the damage an attacker can do within the container, adding an extra layer of defense. Check out this link for more details on Docker security and potential breakout scenarios: https://github.com/ivanversluis/pentest-hacktricks/blob/master/linux-unix/privilege-escalation/docker-breakout.md

YeagerAI-Bulat (commented):
I have modified script to check if the RPC API on port 4000 available every 5 seconds in 60 attempts(300 seconds) and then fail the job

AgustinRamiroDiaz (approved):

cristiam86:
@YeagerAI-Bulat @AgustinRamiroDiaz could you please resolve the conversations that has been addressed?

docker/Dockerfile.frontend
YeagerAI-Bulat: I don't think it's necessary to manually manage the npm cache when using Buildkit. Buildkit automatically leverages layer caching for node modules based on your package-lock.json or yarn.lock file

docker/Dockerfile.genvm
YeagerAI-Bulat: I have changed it

docker/Dockerfile.rpc
YeagerAI-Bulat: +

cristiam86 (approved):

docker/Dockerfile.database
```diff
@@ -1,9 +1,13 @@
-# Use the official PostgreSQL image from the Docker Hub
-FROM postgres:12.17
+# Use the official PostgreSQL image with Alpine linux from the Docker Hub
+FROM postgres:12.19-alpine
```
AgustinRamiroDiaz, line 2: Is this change backwards compatible?

docker/Dockerfile.database
```diff
@@ -1,9 +1,13 @@
...
 
 # Copy the initialization script or SQL file into the container
 # It should be an SQL file or a shell script that runs SQL commands
-COPY ./backend/database_handler/initialization/init-db.sql /docker-entrypoint-initdb.d/
\ No newline at end of file
+<<<<<<< HEAD
+COPY ./database/initialization/init-db.sql /docker-entrypoint-initdb.d/
+=======
+COPY ./backend/database_handler/initialization/init-db.sql /docker-entrypoint-initdb.d/
+>>>>>>> origin/staging
```
AgustinRamiroDiaz, line 13: Please resolve these merge conflicts

YeagerAI-Bulat: Done

docker/Dockerfile.database
AgustinRamiroDiaz: Tested it locally and seems like it's backwards compatible :+1:

AgustinRamiroDiaz (approved):

cristiam86:
@YeagerAI-Bulat please update from staging and solve conflicts. We can merge it after that

cristiam86 (approved):

PR #243, opened 2024-07-01
239 SIM-FE-Create accounts in the frontend with its pair of keys

bradleySuira:
This PR includes a simple implementation to use local generated accounts in the front end using viem library 
- we store now a random private key in `accountsStore.privateKeys` local storage
- we generate the accounts based on the private key stored
- we show the address of the account related to the pk stored


frontend/src/utils/accounts.ts
```diff
@@ -0,0 +1,10 @@
+import { generatePrivateKey, privateKeyToAccount } from 'viem/accounts'
+
+export const getAccount = (privateKey: `0x${string}`) => {
+    return privateKeyToAccount(privateKey)
+}
+
+export const getPrivateKey = () => {
```
cristiam86, line 7: @bradleySuira should these wrappers go on "utils"?
generate looks like a better name for this one and getAccountFromPrivatekey also looks better in my opinion for the other method

frontend/src/views/ProfileView.vue
```diff
@@ -31,9 +31,9 @@ const handleShowSetDefaultAccount = (address: string) => {
 const handleHideSetDefaultAccount = (address: string) => {
   showSetDefaultAccount.value[address] = false
 }
-const setCurentUserAddress = (address: string) => {
+const setCurentUserAddress = (address: `0x${string}`) => {
   if (address) {
-    store.currentUserAddress = address
+    store.currentPrivateKey = address
```
cristiam86, line 36: @bradleySuira, how are you assigning an address to a private key?

cristiam86 (changes requested):
Would you also remove the create account endpoint from the backend?

frontend/src/utils/accounts.ts
bradleySuira: for now in utils but we can define later if we need more related to wallet restore and similar the name is now your suggestion your are right

frontend/src/views/ProfileView.vue
bradleySuira: done, that's because we used address in the past

bradleySuira:
> Would you also remove the create account endpoint from the backend? Also, please wait until the refactor is merged and change the destination of this PR to staging

for now I removed the request from the frontend and json rpc service, in the backend we have some dependencies at least in tests that needs the create account call for example when fund accounts, maybe we need to define how to create accounts in the same format like in frontend but in the backend then can be removed completely

cristiam86:
@bradleySuira please update from staging and solve conflicts

tests/common/response.py
```diff
@@ -1,4 +1,6 @@
 def assert_dict_struct(data, structure):
+    print("data", data)
```
cristiam86, line 2: @bradleySuira please remove this prints :)

cristiam86 (changes requested):

tests/common/response.py
bradleySuira: fixed

cristiam86 (approved):

PR #253, opened 2024-07-03
233 SIM-problem parsing node configuration

bradleySuira:
This PR includes minor fixes for validators configuration:

- Validate config parse 
- Validate input number only integers for stake


frontend/src/views/Simulator/SettingsView.vue
```diff
@@ -77,6 +77,16 @@ async function handleDeleteValidator() {
   }
 }
 
+const handleNumberInput = (event: Event) => {
+  const formattedValue = parseInt((event?.target as any)?.value || '', 10);
+
+  if (!isNaN(formattedValue)) {
+    (event?.target as any).value = '';
```
cristiam86, line 84: @bradleySuira please add some explanation to this. Why are you assigning different values in two consecutive lines?

cristiam86 (changes requested):

frontend/src/views/Simulator/SettingsView.vue
bradleySuira: sure that's because the browser behavior 
- if you sed the .value with the updated value that we want, the cursor position will be at the start of the input 
- now if we put an empty string then set the value the cursor is at the end of the value 

for example 
- when you set the value directly and the value if the input is **1000** the cursor will be at the begging **[**1000
- if we do that we have the cursor will be at the end  1000**]**

AgustinRamiroDiaz (approved):
I'm now able to create validators :+1:

PR #262, opened 2024-07-08
225 SIM-Improve error feedback on contract code changes

bradleySuira:


backend/node/genvm/base.py
AgustinRamiroDiaz, line 1: Please use the python black formatter
https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter

backend/node/genvm/base.py
```diff
@@ -133,9 +134,59 @@ async def run_contract(
...
 
         namespace = {}
-        exec(contract_code, globals(), namespace)
+        try:
+            exec(contract_code, globals(), namespace)
+        except SyntaxError as err:
+            cl, exc, tb = sys.exc_info()
+            
+            return {
+                "exec_error": {
```
AgustinRamiroDiaz, line 143: nit: I think it would be better to create some type for this function which could encapsulate both the happy path and the errors raised

backend/node/genvm/base.py
```diff
@@ -133,9 +134,59 @@ async def run_contract(
     def get_contract_schema(contract_code: str) -> dict:
 
         namespace = {}
-        exec(contract_code, globals(), namespace)
+        try:
+            exec(contract_code, globals(), namespace)
+        except SyntaxError as err:
```
AgustinRamiroDiaz, line 139: Exception handling looks great :100: 

Could you provide some docs as comments? I mean around how to get these types of errors and the `sys` module
The part with the extract of information is very cryptic at first glance.

backend/node/genvm/base.py
cristiam86: @bradleySuira, Also, the return schema of the function should stay the same, adding the errors. Let's move these checks to the code_enforcement and run it like you do on line 189. If code_enforcement returns an error, you can finish this function execution earlier, returning whatever you can return (class_name, etc) plus the errors:

```
contract_schema = {
            "class": class_name,
            "methods": methods,
            "variables": variables,
            "errors": []
        }
```

cristiam86:
Fixed in https://github.com/yeagerai/genlayer-simulator/pull/382

PR #263, opened 2024-07-10
SIM-Pin ollama version

AgustinRamiroDiaz:
# Why

Using `latest`:
- is not reproducible, which can lead to unexpected errors
- takes time to connect to dockerhub and check for new images

I think we should manually upgrade when necessary

# Motive

Downloading newer ollama images was slowing my dev flow

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,7 +50,7 @@ services:
       - "${WEBREQUESTPORT}"
 
   ollama:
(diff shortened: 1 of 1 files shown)


cristiam86 (approved):

cristiam86:
@AgustinRamiroDiaz, before merging this one, could you please create an issue in the GH project and change the title accordingly?

PR #273, opened 2024-07-11
272 SIM - BE - Fix - json load for validator config 

bradleySuira:
This pr includes a hot fix for validator config parsing


AgustinRamiroDiaz (changes requested):
We've talked with Bradley outside, and I don't think this is the right way to tackle this issue. I'll put more information on the issue

bradleySuira:
> We've talked with Bradley outside, and I don't think this is the right way to tackle this issue. I'll put more information on the issue

pushed the fixes and updated the ticket with the details but want to add here

- As you said be were storing a string in a jsonb field in database
- I ended sending when call to create and update methods a valid json from the frontend
- in the backend side in both, update and create, I ended using this ` "config": AsIs("'" + json.dumps(config) + "'"),`, with this change the field is stored as a json in the database and not as string
- Because the last point, we don't need to cal anymore the json.loads() method when we get the validators data, so `json.loads(validator_data["config"]),`  is now `validator_data["config"]` like we have before

I tested addling listing and updating the validators and also the contract deployment and methods calls

backend/database_handler/validators_registry.py
```diff
@@ -57,7 +57,7 @@ def create_validator(
             "stake": stake,
             "provider": provider,
             "model": model,
-            "config": json.dumps(config),
+            "config": AsIs("'" + json.dumps(config) + "'"),
```
AgustinRamiroDiaz, line 60: @bradleySuira why can't we just directly use
```suggestion
            "config": config,
```
? From [the docs](https://www.psycopg.org/docs/extras.html#adapt-json) seems like it automatically adapts it out of the box. If not, it suggests to use the [Json adapter](https://www.psycopg.org/docs/extras.html#psycopg2.extras.Json)

AgustinRamiroDiaz (approved):

backend/database_handler/validators_registry.py
bradleySuira: thanks @AgustinRamiroDiaz used at the end `from psycopg2.extras import Json` and works well

cristiam86 (approved):

PR #284, opened 2024-07-16
280 sim contractrunner from address should be accessed with dot notation

cristiam86:
# What
The contract_runner helper has been refactored to a class so developers can access its attributes such as `from_address` with dot notation instead of dict notation (square brackets and double quotes)


backend/node/genvm/base.py
```diff
@@ -13,6 +13,16 @@
 from backend.node.genvm.code_enforcement import code_enforcement_check
 
 
+class ContractRunner:
+    def __init__(self, mode: str, node_config: dict):
+        self.mode = mode
```
AgustinRamiroDiaz, line 18: I'm still not familiar entirely with the codebase, so just to learn: Would it be better to have 2 types of `ContractRunners`, one for validators and another one for leaders?

backend/node/genvm/base.py
```diff
@@ -13,6 +13,16 @@
 from backend.node.genvm.code_enforcement import code_enforcement_check
 
 
+class ContractRunner:
```
AgustinRamiroDiaz, line 16: Could you add comments explaining what each field represents please?

backend/node/genvm/base.py
```diff
@@ -13,6 +13,16 @@
...
+    def __init__(self, mode: str, node_config: dict):
+        self.mode = mode
+        self.node_config = node_config
+        self.from_address = None
```
AgustinRamiroDiaz, line 20: when is `None` a valid value for `from_address`?

AgustinRamiroDiaz (approved):
Changes look good and issue is addressed :+1:

backend/node/genvm/base.py
cristiam86: I don't think so :) The ContractRunner supports the execution of the IC code. The mode is just a property to understand how to execute some parts of the code and how to process the output of the execution.

backend/node/genvm/base.py
```diff
@@ -13,6 +13,16 @@
...
+        self.gas_used = 0
+        self.eq_num = 0
+        self.eq_outputs = {"leader": {}}
```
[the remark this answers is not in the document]
cristiam86: Yeah! let's talk about this later please it wasn't clear for me to add another package only for this and how to set the default value for the eq_outputs field

backend/node/genvm/base.py
cristiam86: Done!

backend/node/genvm/base.py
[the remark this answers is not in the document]
AgustinRamiroDiaz: No problem, just to be clear it's from the standard python's library

AgustinRamiroDiaz (approved):

backend/node/genvm/base.py
cristiam86: sorry, this was unanswered: when you need to get a contract schema or read from a contract state, you are not sending a transaction thus the from_address is None.
More info: https://stackoverflow.com/a/51850115/637623

PR #292, opened 2024-07-17
291-sim-propose-a-pr-template

### LINKED ISSUE #291: SIM-Propose a PR template

AgustinRamiroDiaz:
Fixes #291

Signed-off-by: Agustín Ramiro Díaz <agustin.ramiro.diaz@gmail.com>

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -37,3 +37,7 @@ Fixes #issue-number-here
 # Reviewing tips
 
 <!-- What can you tell the reviewer to make the review easier? -->
+
+# User facing release notes
+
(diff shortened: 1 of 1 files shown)


cristiam86:
@AgustinRamiroDiaz this is great. Do you think that we could include a section for the developer to explain this PR in a way they would do it to communicate it to outsiders? In the best case, this could be copy&pasted into a Tweet

AgustinRamiroDiaz:
> @AgustinRamiroDiaz this is great. Do you think that we could include a section for the developer to explain this PR in a way they would do it to communicate it to outsiders? In the best case, this could be copy&pasted into a Tweet

Sounds good! Added :D

cristiam86 (approved):

PR #299, opened 2024-07-17
274 SIM-Improvements in the getters section of the UI 

bradleySuira:
This PR includes:
- Refactor to the component for Execute transactions to be reusable for getter methods
- update Run and debug screen to reuse same component for read and write methods


frontend/src/components/Simulator/ExecuteContractMethods.vue
```diff
@@ -31,11 +32,16 @@ const inputs = computed<{ [k: string]: any }>(() => {
...
 })
 
-
 watch(() => props.abi, (newValue) => {
   method.value = undefined
-  methodList.value =  Object.entries(newValue?.methods || {})
-    .filter((m) => !m[0].startsWith('_') && !m[0].startsWith('get_'))
+  methodList.value = Object.entries(newValue?.methods || {})
+    .filter((m) => {
+      if (props.mode === 'read') {
```
cristiam86, line 39:
```suggestion
    const isGetterMethod = m[0].startsWith('get_')
    const isPrivateMethod = m[0].startsWith('_')
    
    return !isPrivateMethod && (props.mode === 'read' ? isGetterMethod : !isGetterMethod)
```

cristiam86 (changes requested):
I see value in trying to refactor some code and make some improvements but adding that in a PR that was meant to be quick and easy makes the review slower and the change riskier.
Also, I'm not sure that reusing the same component for the read and write methods clarify things. It seems that the logic has only gone to another place.

frontend/src/components/Simulator/ExecuteContractMethods.vue
bradleySuira: @cristiam86 , in this case you mean that is better to use a single component for both and just render the two sections ?

cristiam86: Probably yes but I wont revert those changes now that they have been made

bradleySuira: @cristiam86 fixed now I used the `ExecuteContractMethods` as a container and then Another component that list the methods and we separated by getters and writters 

About the fixes and refactor will take in account for next situation and create a separated PR with fixes before the new feature.

cristiam86:
I see an improvement in the UI of the getter methods, but we need to improve it.
1. The current account section is taking up too much space. It can be placed at the top of the deploy section as a global selector to set the account for deploying and calling the contract methods. The important part is that it doesn't take up too much space.
2. The Latest transactions section is very hidden; for now, let's make the deployment section collapsible (opened by default).

bradleySuira:
> I see an improvement in the UI of the getter methods, but we need to improve it.
> 
> 1. The current account section is taking up too much space. It can be placed at the top of the deploy section as a global selector to set the account for deploying and calling the contract methods. The important part is that it doesn't take up too much space.
> 2. The Latest transactions section is very hidden; for now, let's make the deployment section collapsible (opened by default).

Thanks @cristiam86 added the changes you requested, this is an screenshot to check how it looks opened by default 
Screenshot 2024-07-18 at 2 35 56 PM

cristiam86:
> > I see an improvement in the UI of the getter methods, but we need to improve it.
> > 
> > 1. The current account section is taking up too much space. It can be placed at the top of the deploy section as a global selector to set the account for deploying and calling the contract methods. The important part is that it doesn't take up too much space.
> > 2. The Latest transactions section is very hidden; for now, let's make the deployment section collapsible (opened by default).
> 
> Thanks @cristiam86 added the changes you requested, this is an screenshot to check how it looks opened by default Screenshot 2024-07-18 at 2 35 56 PM

Great, but the Current Account selector should be outside the Deploy section since it is being used for deploying and calling the contract.

cristiam86:
@bradleySuira the layout looks better but I can't see any "Read" or "Execute" methods in any contract:
Screenshot 2024-07-23 at 15 45 06

cristiam86:
Fixed in #362

Issue #306, opened 2024-07-19
SIM-BE-Use SQLAlchemy instead of handcrafted SQL Queries


AgustinRamiroDiaz:
# Why

- SQLAlchemy provides higher level of abstractions, which will be easier to use and understand
- SQLAlchemy provides better type safety, which is less error prone

# Examples
- https://github.com/yeagerai/genlayer-simulator/pull/307

PR #307, opened 2024-07-19
SIM-BE-Use SQLAlchemy instead of handcrafted SQL Queries

AgustinRamiroDiaz:
Part of #306

# What

Uses SQLAlchemy engine and session in validator retrieval from db

# Why

- to set an example on how we can improve the db handler

# Testing done

- ran e2e tests

# Decisions made

This is a small example, the objective is to slowly move towards this pattern

# Checks

- [x] I have tested this code

# Reviewing tips

Check out the docs http://docs.sqlalchemy.org/

### DIFF
--- a/backend/database_handler/validators_registry.py
+++ b/backend/database_handler/validators_registry.py
@@ -10,6 +10,7 @@ from backend.errors.errors import ValidatorNotFound
 from sqlalchemy.orm import Session
 
 
+# the to_dict function lives in this module and not in models.py because it's on this layer of abstraction where we convert database objects to our custom data structures
 def to_dict(validator: models.Validators) -> dict:
     return {
         "id": validator.id,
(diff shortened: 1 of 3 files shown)


backend/database_handler/validators_registry.py
```diff
@@ -25,14 +40,17 @@ def _parse_validator_data(self, validator_data: dict) -> dict:
...
+            validator_data = (
+                session.query(models.Validators)
+                .filter_by(address=validator_address)
+                .first()
+            )
 
         if not validator_data:
             raise ValidatorNotFound(validator_address)
-        return self._parse_validator_data(validator_data)
+        return validator_data.to_dict()
```
cristiam86, line 53: @AgustinRamiroDiaz, is this to_dict method the one defined in the models? If so, the one defined in this file is not currently used. Are we keeping it for the future?

AgustinRamiroDiaz: @cristiam86 you are right! Fixed

cristiam86 (approved):

PR #310, opened 2024-07-19
SIM-BE-Decouple configuration

### LINKED ISSUE #309: SIM-BE-Decouple configuration
# What

- improve configuration by creating a configuration struct,
- all IO should be only accessed on top level module. This includes envvars, files, networking
  - This aligns with [hexagonal architecture](https://en.wikipedia.org/wiki/Hexagonal_architecture_(software)) or [onion architecture](https://www.codeguru.com/csharp/understanding-onion-architecture/)

# Why

- Easier testing with dependency injection
- Clearer configuration by defining dependencies
- More versatile modules: for example, a module which relies on envvars or files is harder to configure and customize than a module which can be configured through function parameters

# Examples

AgustinRamiroDiaz:
Fixes #309

# What

- adds global configuration struct in top module
- adds `get_ollama_url` as configuration to required functions
  - this also serves as dependency injection, functions behave as interfaces (some call it the [strategy pattern](https://fsharpforfunandprofit.com/posts/convenience-functions-as-interfaces/))

# Why

described in issue

# Testing done

- I've ran e2e tests locally

# Decisions made

Only example demonstration, no big refactor

# Checks

- [x] I have tested this code

### DIFF
--- a/.github/workflows/backend_integration_tests_pr.yml
+++ b/.github/workflows/backend_integration_tests_pr.yml
@@ -41,10 +41,14 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-buildx-
 
+      - name: Build Docker image
+        run: docker compose build
+
       - name: Run Docker Compose
         run: docker compose up -d
 
       - name: Wait for services to be up
+        timeout-minutes: 5
         run: |
           while ! curl -X POST -H "Content-Type: application/json" -d '{"jsonrpc":"2.0","method":"ping","params":[],"id":1}' http://0.0.0.0:4000/api > /dev/null; do
             echo "Waiting for rpc server..."
--- a/.github/workflows/backend_integration_tests_pr.yml
+++ b/.github/workflows/backend_integration_tests_pr.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   test:
-    if: ${{ github.event.review.state == 'APPROVED' || (github.event_name == 'pull_request' && github.base_ref == 'staging') }}
+    if: ${{ github.event.review.state == 'APPROVED' || (github.event_name == 'pull_request' && github.base_ref == 'staging') || github.event_name == 'workflow_dispatch' }}
     runs-on: ubuntu-latest
 
     env:
--- a/backend/database_handler/validators_registry.py
+++ b/backend/database_handler/validators_registry.py
@@ -71,7 +71,7 @@ class ValidatorsRegistry:
         model: str,
         config: json,
     ):
-        validator = self._get_validator_or_fail(validator_address)
+        self._get_validator_or_fail(validator_address)
 
         update_condition = f"address = '{validator_address}'"
         validator = {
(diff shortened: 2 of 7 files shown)


backend/database_handler/validators_registry.py
```diff
@@ -71,7 +71,7 @@ def update_validator(
         model: str,
         config: json,
     ):
-        validator = self._get_validator_or_fail(validator_address)
```
AgustinRamiroDiaz, line 74: extra: this value was not being used

backend/node/genvm/llms.py
```diff
@@ -93,7 +93,7 @@ async def call_heuristai(
         # raise Exception(chunk.json(), dir(chunk), chunk.choices[0].delta.content)
         try:
             output += chunk.choices[0].delta.content
-        except Exception as e:
```
AgustinRamiroDiaz, line 96: extra: this value was not being used

backend/protocol_rpc/endpoints.py
```diff
@@ -4,7 +4,7 @@
 import json
 from functools import partial
 from flask_jsonrpc import JSONRPC
-from flask import Flask
```
AgustinRamiroDiaz, line 7: extra: this value was not being used

cristiam86 (approved):

.github/workflows/backend_integration_tests_pr.yml
```diff
@@ -41,10 +41,14 @@ jobs:
...
             ${{ runner.os }}-buildx-
 
+      - name: Build Docker image
+        run: docker compose build
+
       - name: Run Docker Compose
         run: docker compose up -d
 
       - name: Wait for services to be up
+        timeout-minutes: 5
```
AgustinRamiroDiaz, line 51: Extra: tests were failing after 6 hours, so I've added a timeout to this infinite loop

.github/workflows/backend_integration_tests_pr.yml
```diff
@@ -12,7 +12,7 @@ on:
 
 jobs:
   test:
-    if: ${{ github.event.review.state == 'APPROVED' || (github.event_name == 'pull_request' && github.base_ref == 'staging') }}
+    if: ${{ github.event.review.state == 'APPROVED' || (github.event_name == 'pull_request' && github.base_ref == 'staging') || github.event_name == 'workflow_dispatch' }}
```
AgustinRamiroDiaz, line 15: extra: in order to check if the tests are passing I've added this line to be able to call them manually

AgustinRamiroDiaz: Test passed at https://github.com/yeagerai/genlayer-simulator/actions/runs/10093412255

cristiam86 (approved):
I like this approach but something that bugs me a little is seeing that get_ollama_url being propagated from functions to functions a lot of times.

Do you have any idea on how we could solve that? Should we pass the whole config until a certain value is needed?

Issue #211, opened 2024-06-07
SIM-OPS-Generate PR from staging to main with changelog


cristiam86:
Generate a GH Action that produces the release PR (Staging to Main), adding every PR title and generating the change log

PR #335, opened 2024-07-25
211 SIM-OPS-Generate PR from staging to main with changelog

AgustinRamiroDiaz:
Fixes https://github.com/yeagerai/genlayer-simulator/issues/211

# What

- Creates a PR from branches of the form `release-<semver>`, like for example `release-v1.2.3`

# Why

- to automate releases

# Testing done

I've created the branch `release-test` from this branch and
- the GH Action has run successfully https://github.com/yeagerai/genlayer-simulator/actions/runs/10113269538/job/27969244092
- the PR was created successfully https://github.com/yeagerai/genlayer-simulator/pull/346

# Decisions made

I decided to go with branch push as trigger instead of `workflow_dispatch` because I feel it's more comfortable, but I can change that if you think it's more comfortable to do it in the UI

# Checks

- [ ] I have tested this code

# Reviewing tips

GitHub API and CLI docs are very useful


.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,64 @@
...
+          echo $latest_release
+          echo "tag=$latest_release" >> "$GITHUB_OUTPUT"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Generate release notes
+        id: generate_release_notes
+        # https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#generate-release-notes-content-for-a-release
+        run: |
+          response=$(curl -L \
```
YeagerAI-Bulat, line 43: What if this curl command will fail from some rate-limiting rule? How to handle such errors?

.github/workflows/release-pr.yml
```diff
@@ -0,0 +1,64 @@
...
+            -X POST \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/repos/yeagerai/genlayer-simulator/releases/generate-notes \
+            -d '{"tag_name":"${{ steps.extract_tag.outputs.tag }}","target_commitish":"${{ steps.extract_branch.outputs.branch }}","previous_tag_name":"${{ steps.get_latest_release.outputs.tag}}"}' )
+          echo $response
+
```
YeagerAI-Bulat, line 51: Can you use https://github.com/marketplace/actions/release-notes-generator instead of this?

YeagerAI-Bulat (changes requested):

.github/workflows/release-pr.yml
AgustinRamiroDiaz: First time I see that one, but I've researched a lot about tools to automate this and I didn't find anyone that would satisfy our needs. All of them had little to no flexibility to customize them.

For example, the one you are sharing here only works with "milestones"

Do you know how to use it for this use case?
We need to create release notes when there's a release branch created, and add those notes to a PR body

.github/workflows/release-pr.yml
AgustinRamiroDiaz: This hasn't failed for me in my tests. I think for now we can go forward by just retrying the action manually if it fails and addressing it in a fix PR. I don't think we should complicate it now :)

cristiam86 (approved):

PR #378, opened 2024-08-13
376 sim allow to disable starting and success logs on rpc endpoints

### LINKED ISSUE #376: SIM-Allow to disable starting and success logs on RPC endpoints

cristiam86:
Fixes #376 

# What

- A list of endpoints to disable the login at transaction "Starting" and "Successful execution" has been added in the endpoints_generator.
- Some types were added to continue this general refactor
- General logs for flask app and socketio has been disable

# Why

- The terminal is getting swamped with logs. It is hard for a developer to find something there since there are too many logs.

# Testing done

- added the following endpoints to the disable list: "get_transaction_by_id", "get_contract_schema_for_code",  and "get_contract_schema". Checked that those endpoints doesn't print the starting or success execution logs.
- checked that other same logs for other endpoints are getting printed
- backend integration tests ran successfully

# Decisions made

Adding this array of disabled info logs as a mechanism to add/remove endpoints where the dev wants to see the logs.

# Checks

- [x] I have tested this code

# Reviewing tips

Try adding/removing endpoints in the `disableInfoLogs` to check if those endpoints are logged or not for Starting and successful execution.

# User facing release notes

Logging has been improved to clean the terminal and allow the developers to find what they are looking for there.


backend/protocol_rpc/server.py
```diff
@@ -75,6 +75,7 @@ def run_socketio():
         host="0.0.0.0",
         allow_unsafe_werkzeug=True,
     )
+    logging.getLogger("werkzeug").setLevel(logging.ERROR)
```
AgustinRamiroDiaz, line 78: can we set this only if there's an envvar like "DEBUG" or "DEV"? These logs might have important warnings when running in production

backend/protocol_rpc/endpoint_generator.py
```diff
@@ -1,24 +1,54 @@
...
+
+"""
+Add the function name to this list to disable info logs for that endpoint
+The info logs are the Starting... and Endpoint successfully executed messages
+"""
+disableInfoLogs = [
+    "get_transaction_by_id",
+    "get_contract_schema_for_code",
+    "get_contract_schema",
+]
```
AgustinRamiroDiaz, line 17: nit: it'd be great if we could configure this through envvars. We can encode this as a JSON in an envvar and parse it here

AgustinRamiroDiaz (approved):

backend/protocol_rpc/server.py
cristiam86: Done!

backend/protocol_rpc/endpoint_generator.py
cristiam86: Done!

denishacquin (approved):

Issue #389, opened 2024-08-16
SIM-Show contract prints in the UI console


cristiam86:

cristiam86:
Fixed in #386

Issue #386, opened 2024-08-15
SIM-Improve logging messages during transaction execution workflow


cristiam86:

denishacquin:
Any chance I could have a little more details on this please?

cristiam86:
This is a duplicate of #436

PR #390, opened 2024-08-16
386 sim improve logging messages during transaction execution workflow

cristiam86:
Fixes: 
- #386
- #389

# What

- Passed the msg_handler to the consensus, node, and genvm for its use
- Captured the standard output on the execution of the intelligent contracts to show those in the UI logs
- Small fix in the frontend regarding vertical scroll
- Removed the address validation in the update validation method to prevent it from failing
# Why

- It is a feature very requested that adds important information to the development process

# Testing done

- Manually tested the new feature
- Ran frontend and backend tests

# Decisions made

# Checks

- [x] I have tested this code

# Reviewing tips

Add some prints to your intelligent contracts, execute them, and see those prints in the UI logs

# User facing release notes

You can now see the prints that you add to your intelligent contracts getting logged in the UI logs section.

### DIFF
--- a/backend/node/base.py
+++ b/backend/node/base.py
@@ -5,6 +5,7 @@ from typing import Optional
 from backend.node.genvm.base import GenVM
 from backend.database_handler.contract_snapshot import ContractSnapshot
 from backend.node.genvm.types import Receipt, ExecutionMode, Vote
+from backend.protocol_rpc.message_handler.base import MessageHandler
 
 
 class Node:
@@ -18,6 +19,7 @@ class Node:
         model: str,
         config: dict,
         leader_receipt: Optional[Receipt] = None,
+        msg_handler: MessageHandler = None,
     ):
         self.validator_mode = validator_mode
         self.address = address
@@ -28,7 +30,9 @@ class Node:
             "stake": stake,
         }
         self.leader_receipt = leader_receipt
-        self.genvm = GenVM(contract_snapshot, self.validator_mode, self.validator_info)
+        self.genvm = GenVM(
+            contract_snapshot, self.validator_mode, self.validator_info, msg_handler
+        )
 
     async def exec_transaction(self, transaction: dict):
         transaction_data = transaction["data"]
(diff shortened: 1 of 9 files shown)


backend/node/base.py
```diff
@@ -18,6 +19,7 @@ def __init__(
         model: str,
         config: dict,
         leader_receipt: Optional[Receipt] = None,
+        msg_handler: MessageHandler = None,
```
AgustinRamiroDiaz, line 22: The GenVM is not handling the possibility of `msg_handler == None`, so I would make it a required parameter

backend/node/genvm/base.py
```diff
@@ -141,7 +157,10 @@ async def run_contract(
         execution_result = ExecutionResultStatus.SUCCESS
         error = None
 
-        with safe_globals():
+        # Buffers to capture stdout and stderr
+        stdout_buffer = io.StringIO()
+
+        with redirect_stdout(stdout_buffer), safe_globals():
```
AgustinRamiroDiaz, line 163: should we also redirect stderr?

backend/protocol_rpc/endpoints.py
```diff
@@ -202,8 +208,9 @@ def update_validator(
     model: str,
     config: json,
 ) -> dict:
-    if not accounts_manager.is_valid_address(validator_address):
-        raise InvalidAddressError(validator_address)
+    # Remove validation while adding migration to update the db address
+    # if not accounts_manager.is_valid_address(validator_address):
+    #     raise InvalidAddressError(validator_address)
```
AgustinRamiroDiaz, line 213: should we revert these comments?

backend/protocol_rpc/endpoints.py
```diff
@@ -214,8 +221,9 @@ def delete_validator(
     accounts_manager: AccountsManager,
     validator_address: str,
 ) -> dict:
-    if not accounts_manager.is_valid_address(validator_address):
-        raise InvalidAddressError(validator_address)
+    # Remove validation while adding migration to update the db address
+    # if not accounts_manager.is_valid_address(validator_address):
+    #     raise InvalidAddressError(validator_address)
```
AgustinRamiroDiaz, line 226: should we revert these comments?

backend/node/genvm/base.py
cristiam86: I tried but even the errors were redirected to the stdout. Probably because they are handled by the Exception and printed.

backend/protocol_rpc/endpoints.py
cristiam86: I added them now to allow the validators update. Maybe my comment is not accurate but I meant that I didn't want to add the validation until we migrate the validators address in the DB to be correct

Issue #351, opened 2024-07-29
SIM-DB-Decouple startup script in order to use plain postgres DB


AgustinRamiroDiaz:
# Why

Using plain postgres DB allows for using DBaaS

We do very minimal configurations to the postgres DB which are not necessary given that we have now the migration component.

# Acceptance criteria

Allow deploying the app without custom postgres image

PR #410, opened 2024-08-21
chore: use postgresql public image and add necessary alembic migration

AgustinRamiroDiaz:
Fixes https://github.com/yeagerai/genlayer-simulator/issues/351

# What

- added init alembic migration which creates tables
- removed custom database image

# Why

- for easier deployment, many cloud providers offer managed PostgreSQL, and it's easier to just use the default image

# Testing done

docker compose down + docker compose up + smoke tested the UI

# Decisions made

I've decided to create a new migration and put it as the first one. This is a BREAKING CHANGE that requires re-creating the database (docker compose down + docker compose up)

# Checks

- [ ] I have tested this code

# User facing release notes

Easier deployment by using default `Postgres` image in the database. This change requires re-creating the database with `docker compose down`


.github/workflows/docker-build-and-push-image.yml
```diff
@@ -20,9 +20,6 @@ on:
     secrets:
       dockerhub_token:
         required: true
-  push:
-    tags:
-      - "v*.*.*"
```
AgustinRamiroDiaz, line 25: Extra: this trigger was wrong, since we only want to trigger this by workflow calls
example error https://github.com/yeagerai/genlayer-simulator/actions/runs/10490636510/job/29057821846#step:3:7

docker-compose.yml
```diff
@@ -90,3 +88,7 @@ services:
     depends_on:
       postgres:
         condition: service_healthy
+
+networks:
```
cristiam86, line 92: @AgustinRamiroDiaz should we have this reference to feed_buzz here? Can we maybe find a more generic name?

AgustinRamiroDiaz: My bad, this shouldn't have made it into the PR. I'll remove it

PR #419, opened 2024-08-27
refactor: refactor services and clients

### LINKED ISSUE #408: SIM-FE-Refactor services and clients
This should be done mainly for external libraries and clients such as Viem and the Simulator's backend
The way it should be architected is following the following pattern
UI Components -> Hooks -> Services -> Clients. 
We should enforce unit testing mostly on Hooks and Services.

Also, please take this opportunity to add unit tests.

### LINKED ISSUE #412: SIM-FE-Bug: all contracts get deleted on reset storage, including user-created ones
According to the reset storage modal, it shouldn't delete the contracts that are not examples.

This is due to a discrepency between two checks that don't have the same logic; we need to normalize it to fix the bug as this can be very inconvenient because we make the user loose all their work.

Image

Image

Image

### LINKED ISSUE #413: SIM-FE-Bug: all sample contracts get duplicated when they all have a change
This is due to some wrong conditional checks.

To reproduce, make a change in all files, then reload.

Image

denishacquin:
Fixes #408 
Fixes #413 
Fixes #412 

# What

- Refactored all wallet-related logic in one dedicated hook
- Removed 'provide' for rpc client and wrapped rpc client in a hook instead
- Moved all utils to hooks
- Updated signTransaction unit test
- Fixed a couple of bugs with example contracts / storage reset

# Why

- to normalize and refactor functionality
- to fix a couple of bugs

# Testing done

- lint OK
- e2e OK
- units OK
- tested the bug fixes

# Checks

- [x] I have tested this code

# Tips for reviewing

- Might want to check the refactor of useWallet hook to validate decisions there.

# User facing release notes

- Fixed a couple of bugs with storage resetting


frontend/src/hooks/useShortAddress.ts
```diff
@@ -0,0 +1,14 @@
+export const useShortAddress = () => {
+  function shortenAddress(address?: string) {
+    const prefix = address?.startsWith('0x') ? '0x' : '';
```
cristiam86, line 3: Maybe this is a bit more readable, what do you think? 
```suggestion
    const prefix = '0x';
    const addressWithoutPrefix = address?.replace('0x', '')
    return `${prefix}${address.slice(0,4)}...${address.slice(-4)}
```

frontend/src/hooks/useShortAddress.ts
```diff
@@ -0,0 +1,14 @@
+export const useShortAddress = () => {
```
cristiam86, line 1: @denishacquin, where do we decide to create a new hook just to shorten an address, or do we decide to include this method in the wallet hook?

frontend/src/hooks/useRpcClient.ts
```diff
@@ -0,0 +1,6 @@
+import { RpcClient } from '@/clients/rpc';
+import { JsonRpcService } from '@/services/JsonRpcService';
+
+export function useRpcClient() {
```
cristiam86, line 4: @denishacquin let's do the same for viem please. We should create a /client/web3 that exposes all viem methods that we need and then a /hooks/useWeb3Client

frontend/src/hooks/usePythonSyntax.ts
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable no-useless-escape */
 import * as monaco from 'monaco-editor/esm/vs/editor/editor.api';
 
-export const pythonSyntaxDefinition: monaco.languages.IMonarchLanguage = {
+const pythonSyntaxDefinition: monaco.languages.IMonarchLanguage = {
```
cristiam86, line 4: @denishacquin do we really need a hook for something that doesn't dynamically change?

cristiam86 (changes requested):
@denishacquin great work! I love seeing the utils getting nuked! but I wonders if everything should be a hook as I understand that those are mostly about dynamic values

frontend/src/hooks/useShortAddress.ts
denishacquin: was thinking the same, will bundle them together

frontend/src/hooks/useShortAddress.ts
denishacquin: actually already have a simplified version in upstream branch, will cherry pick and you can check

frontend/src/hooks/usePythonSyntax.ts
denishacquin: good point, moving it to a constant

frontend/src/hooks/useRpcClient.ts
denishacquin: Is this kindof what you have in mind? useWallet would be the same as useWeb3Client I think? image

denishacquin: I split the client and the hook, see changes

frontend/src/hooks/useWallet.ts
```diff
@@ -0,0 +1,23 @@
+import { Web3Client } from '@/clients/web3';
+
+export function useWallet() {
+  const web3 = new Web3Client();
```
cristiam86, line 4: @denishacquin great to see this, but let's add here the wrapped functions instead of just returning the web3 object. Otherwise we are not really decoupling the web3 client.
signTransaction, privateKeyToAccount, and generatePrivateKey seems better to be gotten from the useWallet

cristiam86 (changes requested):
great! almost there, just a small detail just to close it in a perfect way :D

cristiam86 (approved):

cristiam86 (approved):

PR #420, opened 2024-08-27
chore(backend): refactor database connector to use sqlalchemy

AgustinRamiroDiaz:
# What

Refactors our database integration to use SQLAlchemy's ORM

# Why

- better type safety
- higher level of abstractions when interacting with the database

# Testing done

- e2e tests
- created e2e tests for the db integrations with SQLAlchemy
  - They are documented with their own Readme

# Decisions made

SQLAlchemy's ORM depends on a `Session` in order to execute queries. 
Injecting the session to each request was easy since there's a package for it called `Flask-SQLAlchemy`
Injecting the session to our cronjobs (`run_crawl_snapshot_loop` and `run_consensus_loop`) was not as straight forward, and needed a bit of refactoring

More context [here](https://docs.sqlalchemy.org/en/20/orm/session_basics.html#when-do-i-construct-a-session-when-do-i-commit-it-and-when-do-i-close-it)

# Checks

- [x] I have tested this code

# Reviewing tips

SQLAlchemy's documentation is great (yet extensive)

# User facing release notes


backend/consensus/base.py
```diff
@@ -97,9 +100,10 @@ async def exec_transaction(
 
         print(" ~ ~ ~ ~ ~ EXECUTING TRANSACTION: ", transaction)
         # Update transaction status
-        self.transactions_processor.update_transaction_status(
-            transaction["id"], TransactionStatus.PROPOSING.value
-        )
+        with self.dbclient.get_session() as session:
```
cristiam86, line 103: @AgustinRamiroDiaz why do we need to pass the session to the `update_transaction_status` method instead of the TransactionsProcessor using the session it created in the __init__?

backend/database_handler/accounts_manager.py
```diff
@@ -1,87 +1,89 @@
...
-        self, db_client: DBClient, transactions_processor: TransactionsProcessor
-    ):
-        self.db_client = db_client
+    def __init__(self, session: Session, transactions_processor: TransactionsProcessor):
+        self.session = session
         self.transactions_processor = transactions_processor
-        self.db_accounts_table = "current_state"
 
-    def _parse_account_data(self, account_data: dict) -> dict:
+    def _parse_account_data(self, account_data: CurrentState) -> dict:
```
cristiam86, line 18: @AgustinRamiroDiaz do we still need this? why can't we just return the same CurrentState object?

backend/database_handler/accounts_manager.py
```diff
@@ -1,87 +1,89 @@
...
-            updated_account_state = {
-                "data": json.dumps({"balance": new_balance}),
-            }
-
-            self.db_client.update(
-                self.db_accounts_table, updated_account_state, update_condition
-            )
+            # Dicts are mutable objects, we need to do something in order for SQLAlchemy to track their changes https://docs.sqlalchemy.org/en/20/orm/extensions/mutable.html
+            # In this case we are copying the dict, updating it and assigning it back to the object, which will trigger the change tracking
+            # I (Agustín Díaz) like this approach better than using `MutableDict` because it's more explicit than using the MutableDict class, and also doesn't leak the object to the rest of the code (which can be dangerous given that it's mutable)
```
cristiam86, line 72: @AgustinRamiroDiaz I'm not sure if I follow, let's discuss this

backend/database_handler/transactions_processor.py
```diff
@@ -1,95 +1,111 @@
...
-            "type": transaction_data["type"],
-            "status": transaction_data["status"],
-            "consensus_data": transaction_data["consensus_data"],
-            "gaslimit": transaction_data["nonce"],
-            "nonce": transaction_data["nonce"],
-            "r": transaction_data["r"],
-            "s": transaction_data["s"],
-            "v": transaction_data["v"],
-            "created_at": transaction_data["created_at"].isoformat(),
+            "id": transaction_data.id,
```
cristiam86, line 16: @AgustinRamiroDiaz is this still needed?

backend/protocol_rpc/endpoints.py
```diff
@@ -255,25 +255,23 @@ def get_transaction_by_id(
 
 def get_contract_state(
     accounts_manager: AccountsManager,
-    msg_handler: MessageHandler,
     contract_address: str,
     method_name: str,
     method_args: list,
 ) -> dict:
-    if not accounts_manager.is_valid_address(contract_address):
+    if not address_is_in_correct_format(contract_address):
```
cristiam86, line 262: @AgustinRamiroDiaz where is this imported?

backend/protocol_rpc/endpoints.py
```diff
@@ -255,25 +255,23 @@ def get_transaction_by_id(
...
         raise InvalidAddressError(contract_address)
 
-    contract_account = accounts_manager.get_account(contract_address)
+    contract_account = accounts_manager.get_account_or_fail(contract_address)
     node = Node(
         contract_snapshot=None,
         address="",
-        validator_mode=ExecutionMode.LEADER,
+        validator_mode="leader",
```
cristiam86, line 269: @AgustinRamiroDiaz why is it better as a hardcoded string rather than an enum?

backend/protocol_rpc/endpoints.py
```diff
@@ -345,6 +343,24 @@ def send_raw_transaction(
     result["transaction_id"] = transaction_id
 
     return result
+    validator_address = create_new_address()
```
cristiam86, line 346: @AgustinRamiroDiaz is this executed?

backend/consensus/base.py
AgustinRamiroDiaz: Since the `TransactionsProcessor` is only created once at the init of `ConsensusAlgorithm` class, the session will not be usable
The session that I have injected in the top `server.py` module is a session automatically created by `Flask-SQLAlchemy` that is bound to each request in the Flask server.
This is related to the problem I'm mentioning in the description around the refactor and how to inject sessions to cronjobs

backend/database_handler/accounts_manager.py
AgustinRamiroDiaz: Sure!
For anyone else reading: sqlalchemy tracks changes in your models, so that when you `session.commit()` it knows which SQL queries it needs to run
The thing is that with mutable objects (like dictionaries) SQLAlchemy is not able to track its changes, given that SQLAlchemy only knows about the reference
Some solutions possible to this are:
a) copy the mutable object and reassign it (what is done in the code): this creates a new reference so SQLAlchemy can track it
b) use SQLAlchemy's `MutableDict`: this is a custom struct from SQLAlchemy which its able to be tracked by changes in its fields. The problem with using this is that if you expose this dictionary to another function/module and they mutate the dictionary, those changes will get reflected in the database. This is dangerous IMO because one could forget about this, expose the `CurrentState` object and other modules could modify its fields, not knowing they are actually modifying the DB

backend/protocol_rpc/endpoints.py
AgustinRamiroDiaz: fixed!

backend/protocol_rpc/endpoints.py
AgustinRamiroDiaz: this was a problem with merge conflicts. Fixed!

backend/protocol_rpc/endpoints.py
AgustinRamiroDiaz: fixed!

backend/consensus/base.py
AgustinRamiroDiaz: changed and now it's injected :+1:

backend/database_handler/accounts_manager.py
AgustinRamiroDiaz: talked async: we'll not use the database models as our domain objects. At this layer, we need to convert to our domain
currently our domain for accounts is `dict`. We'll change that in the future, is out of scope for this PR

backend/protocol_rpc/endpoints.py
```diff
@@ -398,7 +407,7 @@ def register_all_rpc_endpoints(
     register_rpc_endpoint_for_partial(get_validator, validators_registry)
 
     register_rpc_endpoint_for_partial(get_transaction_by_id, transactions_processor)
-    register_rpc_endpoint_for_partial(get_contract_state, accounts_manager, msg_handler)
```
cristiam86, line 401: @AgustinRamiroDiaz could you please confirm that you removed this from here because it isn't being used inside the get_contract_data function?

cristiam86 (approved):

backend/protocol_rpc/endpoints.py
AgustinRamiroDiaz: you were right! fixed

cristiam86 (approved):

Issue #359, opened 2024-08-05
SIM-Refactor how inference providers are added and configured


AgustinRamiroDiaz:
We need to make this more accessible and configurable.
Should it needs to be in the database and modified through some endpoints?

cristiam86:
The main requirements are:
- We need to replace the current approach of having the JSON file with the providers, models, and default values.
- The providers, models, and default configuration should be stored in the database.
- We should provide endpoints for accessing the available providers and models (get_providers_and_models already exists), and to add new/edit/remove providers, models, and default configuration.
- The current process to create random validators should be refactored to use these values.

PR #475, opened 2024-09-05
feat(backend): refactor how inference providers are added and configured

AgustinRamiroDiaz:
Fixes https://github.com/yeagerai/genlayer-simulator/issues/359

# What

Added configuration for LLM providers
We now have a schema defined in https://json-schema.org/, which checks that LLM provider configurations are compliant.

## Expected flow for new providers

1. Extend the JSON schema to add their providers
1. If not compatible with other plugins, add a plugin to `llms.py`
1. add a migration to load these new providers into the database

Step 3 seems to be the hardest for newcomers, since it would require a guide to use Alembic and SQLAlchemy to fill the database. Also, step 3 could not be desired since we'd be modifying provider configurations that users could've tuned and maybe don't want to change.  
I'm open to ideas

# Why

To allow new OpenAI compliant providers to come in and add their own by self registering

# Testing done

- Created e2e tests and unit tests

# Decisions made

- Decided to use json schema since
  - it's a big standard with an ecosystem of tools
  - JSON is widely adopted, like in our JSON-rpc
- I've made a lot of extras along the way with the intent of improving the codebase as I was passing by

# Checks

- [x] I have tested this code

# Reviewing tips

- Don't check git history
- review in order:
  1. json-schema 
  3. tests
  4. code

# User facing release notes

- Refactored inference providers: now LLM providers and models can be extended and configured easily, opening the door for new providers to register.
- Added new OpenAI o1-mini and o1-preview models
- Added plugin to allow using Anthropic (Claude)


.pre-commit-config.yaml
```diff
@@ -13,7 +13,7 @@ repos:
     hooks:
       - id: backend-unit-pytest
         name: backend unit tests with pytest
-        entry: python3 -m pytest backend
+        entry: pytest tests/unit
```
AgustinRamiroDiaz, line 16: extra: moved unit tests

.vscode/settings.json
```diff
@@ -2,7 +2,11 @@
...
-  "python.testing.pytestArgs": ["tests"],
+  "python.testing.pytestArgs": ["tests", "backend"],
   "python.testing.unittestEnabled": false,
-  "python.testing.pytestEnabled": true
+  "python.testing.pytestEnabled": true,
+  "sonarlint.connectedMode.project": {
+    "connectionId": "YeagerAI",
+    "projectKey": "yeagerai_genlayer-simulator"
+  }
 }
```
AgustinRamiroDiaz, line 12: extra: pytest and sonar vscode config

backend/database_handler/alembic.ini
```diff
@@ -13,7 +13,7 @@ script_location = migration
 
 # sys.path path, will be prepended to sys.path if present.
 # defaults to the current working directory.
-prepend_sys_path = .
+prepend_sys_path = ./backend/database_handler
```
AgustinRamiroDiaz, line 16: needed for imports

backend/database_handler/llm_providers.py
```diff
@@ -0,0 +1,76 @@
...
+    def get_all_dict(self) -> list[dict]:
+        return [
+            {
+                "id": provider.id,
+                "provider": provider.provider,
+                "model": provider.model,
+                "config": provider.config,
+            }
+            for provider in self.session.query(LLMProviderDBModel).all()
+        ]
```
AgustinRamiroDiaz, line 36: this is duplicated since `LLMProvider` does not have an ID. I think that it's fine for now, but I'm open to opinions on "should the domain models have IDs?"

backend/database_handler/models.py
```diff
@@ -110,3 +110,22 @@ class Validators(Base):
...
     )
+
+
+class LLMProviderDBModel(Base):
+    __tablename__ = "llm_provider"
+    __table_args__ = (PrimaryKeyConstraint("id", name="llm_provider_pkey"),)
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, init=False)
+    provider: Mapped[str] = mapped_column(String(255))
+    model: Mapped[str] = mapped_column(String(255))
```
AgustinRamiroDiaz, line 121: I did not add this pair of `(provider, model)` as a unique key since I feel that we'd be limiting the users. I think that it makes sense that users can create many provider configurations, and then use that pool to draw from

backend/database_handler/models.py
```diff
@@ -110,3 +110,22 @@ class Validators(Base):
     created_at: Mapped[Optional[datetime.datetime]] = mapped_column(
         DateTime(True), server_default=func.current_timestamp(), init=False
     )
+
+
+class LLMProviderDBModel(Base):
```
AgustinRamiroDiaz, line 115: One question that I'm thinking now is that if this should have any relationship with the validators, or if this is a pool to draw configurations with that could live independently?

backend/protocol_rpc/endpoints.py
```diff
@@ -131,15 +131,41 @@ def get_contract_schema_for_code(
...
-def get_providers_and_models(config: GlobalConfiguration) -> dict:
-    default_config = get_default_config_for_providers_and_nodes()
-    providers = get_providers()
-    providers_and_models = {}
-    for provider in providers:
-        providers_and_models[provider] = get_provider_models(
-            default_config["providers"], provider, config.get_ollama_url
-        )
-    return providers_and_models
+# TODO: these endpoints shouldn't return a `dict`, but I'm getting `TypeError: return type of dict must be a type; got NoneType instead`
```
AgustinRamiroDiaz, line 134: @cristiam86 maybe you know how to fix this?

docker/Dockerfile.backend
```diff
@@ -25,7 +25,7 @@ COPY backend $path/backend
 FROM base AS debug
 RUN pip install --no-cache-dir debugpy watchdog
 USER backend-user
-CMD watchmedo auto-restart --recursive --pattern="*.py" --ignore-patterns="*.pyc;*__pycache__*" -- python -m debugpy --listen 0.0.0.0:${RPCDEBUGPORT} -m flask run -h 0.0.0.0 -p ${FLASK_SERVER_PORT}
+CMD watchmedo auto-restart --no-restart-on-command-exit --recursive --pattern="*.py" --ignore-patterns="*.pyc;*__pycache__*" -- python -m debugpy --listen 0.0.0.0:${RPCDEBUGPORT} -m flask run -h 0.0.0.0 -p ${FLASK_SERVER_PORT}
```
AgustinRamiroDiaz, line 28: extra: doesn't restart when code exits, which most of the time is when there are "compile" errors

backend/database_handler/models.py
cristiam86: The second one

cristiam86 (approved):
I like the approach and it is a big improvement from what we have now. Still, we need to try to document and follow the process of adding a new provider and model ourselves to identify some potential problems that anyone could have. 
What concerns me more is the fact that a new provider needs to be added in different places because we have the default configs, some enums in the schemas, the database entry, etc. 
Maybe the ideal scenario where someone just adds providers and models from the UI is very hard to achieve but we need to try to be closer to that one.

backend/protocol_rpc/endpoints.py
```diff
@@ -150,6 +176,15 @@ def create_validator(
     model: str,
     config: json,
 ) -> dict:
+
+    validate_provider(
+        LLMProvider(
+            provider=provider,
+            model=model,
+            config=config,
```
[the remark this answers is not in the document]
AgustinRamiroDiaz: @cristiam86 this comment got moved and it's not pointing to the code you mention. Could you remind me on which function is this?

tests/integration/test_llm_providers_registry.py
```diff
@@ -0,0 +1,34 @@
...
+    assert has_success_status(reset_result)
+
+    response = post_request_localhost(payload("get_providers_and_models")).json()
+    assert has_success_status(response)
+
+    default_providers = response["result"]["data"]
+    first_default_provider = default_providers[0]
+    last_provider_id = default_providers[-1]["id"]
+
+    # Create a new provider
```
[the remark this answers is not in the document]
AgustinRamiroDiaz: I didn't wanted to add hardcoded values since they'd be changing with the schema, but you are right. Once we have a stable schema these tests shouldn't change often

I'll create another test (I think this one adds it's own logic so I won't remove it)

tests/unit/test_providers.py
```diff
@@ -0,0 +1,9 @@
+from backend.node.create_nodes.providers import (
+    get_default_providers,
+)
+
+
+def test_default_providers_valid():
+    providers = get_default_providers()
+
+    assert len(providers) > 0
```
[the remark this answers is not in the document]
AgustinRamiroDiaz: `get_default_providers` internally verifies that all providers are valid for the schema. Do you think there's something more that we should check)

backend/database_handler/migration/versions/986d9a6b0dda_add_plugin_and_plugin_config_to_.py
```diff
@@ -0,0 +1,91 @@
...
+        provider = validator.provider
+        model = validator.model
+        default_provider = get_default_provider_for(provider=provider, model=model)
+        conn.execute(
+            validators.update()
+            .where(validators.c.id == id)
+            .values(
+                plugin=default_provider.plugin,
+                plugin_config=default_provider.plugin_config,
+                config=default_provider.config,
```
AgustinRamiroDiaz, line 61: Note that we are overriding existing configs. This is because the schema has changed and it's the simplest way. Handling the logic of migrating current configs is possible, but would be a lot of work and it's probably not a requirement

AgustinRamiroDiaz:
@cristiam86 I've finished the PR and it's now ready to be reviewed

Note that I haven't been able to test the Claude Plugin

cristiam86 (approved):
Good job @AgustinRamiroDiaz. I have one concern about the performance adding/updating validators which now takes ~1 second while before it was instant. Do you happen to know why?  Could it be related to the way we load the JSON providers files? could we paralelize that somehow?

AgustinRamiroDiaz:
@cristiam86 you were right! Fixed in my last commit by adding a cache + running it at startup :+1:

cristiam86 (approved):

PR #479, opened 2024-09-06
refactor: move from transaction id to transaction hash

### LINKED ISSUE #451: Refactor transaction id with transaction Hash
The current transaction ID is a postgress auto-incremental ID that should be changed to the transaction HASH.
This also affects the endpoint "get_transaction_by_id" which should be changed to "get_transaction_by_hash"
In the frontend list, the ID can be changed to the transaction hash's last 4 or 6 characters.

denishacquin:
Fixes #451 

# What

**Move frome _id_ to _hash_ in the transaction model.**

- Changed the model and renamed methods, endpoints, etc
- Created a migration to add hash column in the db (for now keeping the id, but we can remove it when we feel comfortable)
- Created a method to generate hashes similarly to eth in transaction_processor
- Had to add nonces for hash uniqueness (for now it's on-the-fly on the backend side, I have a branch with the whole endpoint and frontend nonce management but that could be done in #468 )

image
image

# Why

To get closer to eth spec.

# Testing done

- unit tests passed
- tested migration
- tested feature
- E2E OK

# Checks

- [x] I have tested this code

# Reviewing tips

- Need to particularly pay attention to the **generate hash method** and the **DB migration**

# User facing release notes

- Implemented transaction hashes


backend/protocol_rpc/endpoints.py
```diff
@@ -69,11 +69,30 @@ def fund_account(
...
-
-    transaction_id = transactions_processor.insert_transaction(
+    transaction_hash = transactions_processor.insert_transaction(
         None, account_address, None, amount, 0
     )
-    return {"transaction_id": transaction_id}
+    return {"transaction_hash": transaction_hash}
+
+
+def send_transaction(
```
cristiam86, line 78: @denishacquin this method is no longer needed. Sending value is handled by the send_raw_transaction method

cristiam86 (changes requested):
Very good job, only one endpoint that got restored when it was removed before

cristiam86 (approved):

backend/database_handler/migration/versions/99d3f1bc5a08_add_hash.py
```diff
@@ -0,0 +1,101 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "99d3f1bc5a08"
+down_revision: Union[str, None] = "f9636f013003"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+# This migration adds a hash column to the transactions table (instead of id) and a transaction_hash column to the transactions_audit table (instead of transaction_id).
+# To remain backwards compatible and downgradable, the hash and nonce are set to the current id on existing transactions.
```
AgustinRamiroDiaz, line 23: nit: in the start of the file there's a section for these kinds of comments

backend/database_handler/migration/versions/99d3f1bc5a08_add_hash.py
```diff
@@ -0,0 +1,101 @@
...
+def upgrade() -> None:
+    # Add the new 'hash' column
+    op.add_column(
+        "transactions", sa.Column("hash", sa.String(length=255), nullable=True)
+    )
+
+    # Update the 'hash' column with the current 'id' value
+    op.execute("UPDATE transactions SET hash = CAST(id AS VARCHAR(255))")
```
AgustinRamiroDiaz, line 31: Doesn't this hash need to coincide with the hash logic from `TransactionsProcessor._generate_transaction_hash`?

backend/database_handler/models.py
```diff
@@ -57,10 +57,10 @@ class Transactions(Base):
     __tablename__ = "transactions"
     __table_args__ = (
         CheckConstraint("type = ANY (ARRAY[0, 1, 2])", name="transactions_type_check"),
-        PrimaryKeyConstraint("id", name="transactions_pkey"),
+        PrimaryKeyConstraint("hash", name="transactions_pkey"),
     )
 
-    id: Mapped[int] = mapped_column(Integer, primary_key=True, init=False)
+    hash: Mapped[str] = mapped_column(String(255), primary_key=True, unique=True)
```
AgustinRamiroDiaz, line 63: nit: from my tests, all transaction hashes have 66 ("0x" + 64) characters, it'd be nice to limit that here too

backend/database_handler/models.py
```diff
@@ -91,7 +91,7 @@ class TransactionsAudit(Base):
     __table_args__ = (PrimaryKeyConstraint("id", name="transactions_audit_pkey"),)
 
     id: Mapped[int] = mapped_column(Integer, primary_key=True, init=False)
-    transaction_id: Mapped[Optional[int]] = mapped_column(Integer)
+    transaction_hash: Mapped[Optional[str]] = mapped_column(String(255))
```
AgustinRamiroDiaz, line 94: nit: could you add this as a Foreign Key?

backend/database_handler/transactions_processor.py
```diff
@@ -1,9 +1,13 @@
 # consensus/services/transactions_db_service.py
+import rlp
 
 from .models import Transactions, TransactionsAudit
 from sqlalchemy.orm import Session
 
 from .models import TransactionStatus
+from hashlib import sha3_256
```
AgustinRamiroDiaz, line 8: nit: this is not being used

frontend/src/hooks/useContractQueries.ts
```diff
@@ -100,7 +100,7 @@ export function useContractQueries() {
         const tx: TransactionItem = {
           contractAddress: result?.data.contract_address,
           localContractId: contract.value?.id ?? '',
-          txId: result?.data.transaction_id,
+          hash: result?.data.transaction_hash,
           type: 'deploy',
           status: 'PENDING',
           data: {},
```
AgustinRamiroDiaz, line 106: nit: result is not null in this context
```suggestion
          contractAddress: result.data.contract_address,
          localContractId: contract.value?.id ?? '',
          hash: result.data.transaction_hash,
          type: 'deploy',
          status: 'PENDING',
          data: {},
```

frontend/src/hooks/useDb.ts
```diff
@@ -15,6 +15,13 @@ class GenLayerSimulatorDB extends Dexie {
       transactions:
         '++id, type, status, contractAddress, localContractId, txId',
     });
+
+    this.version(3).stores({
+      contractFiles: 'id',
+      deployedContracts: '[contractId+address]',
+      transactions:
+        '++id, type, status, contractAddress, localContractId, hash',
+    });
```
AgustinRamiroDiaz, line 24: what's this? I'm not knowledgeable on the frontend :D

AgustinRamiroDiaz (approved):
Changes look good! all minor comments

My main question is the one about the transaction hash being the current ID

backend/database_handler/migration/versions/99d3f1bc5a08_add_hash.py
denishacquin: gotcha

frontend/src/hooks/useDb.ts
denishacquin: This is an updated model for the local IndexedDB to tell it that there's now a hash column.

backend/database_handler/models.py
denishacquin: Added foreign key constraint, please review

backend/database_handler/models.py
denishacquin: Replaced 255 by 66 in models: transactions.hash, transaction_audits.transaction_hash, as well as in migration file, please review

backend/database_handler/migration/versions/99d3f1bc5a08_add_hash.py
denishacquin: I replicated the hash generation logic in the migration script, however:
- Had to add some eth packages to the requirements of the db handler, is that OK?
- It effectively generates hash for legacy transactions, but it's not mirrored in the frontend. I played around with updating the legacy TX's in the frontend but it's a bit of a rabbit hole; maybe we can leave the frontend Txs without hashes and if people have issue they can just clear the local storage.

cristiam86: I wouldn't care about transactions hash migration in the frontend. We could do something temporary like showing the ID for the ones that doesn't have hash

denishacquin: I thought the same. What I did was migrating them from **id: 1** to **hash: 0x1** just to populate the UI. There might be edge cases with people who had pending transactions during the upgrade, but I think it would be going too far to handle thoses - worst case they can simply clear their local transactions.

cristiam86 (approved):

Issue #168, opened 2024-05-16
SIM-Re-execute consensus when validators don't agree with the leader


cristiam86:

AgustinRamiroDiaz:
@cristiam86 could you provide current behavior and desired behavior in the description?

cristiam86:
Right now, in /backend/consensus/base.pyL169, when the consensus is not reached, an exception is raised.
Instead, it should select a new leader and execute the transaction again.
If all leaders have been rotated and there is still no consensus, the transaction is marked as undetermined: https://docs.google.com/document/d/1qUkQpR3urVNXJvlucmEfeOQJH510j7_XOqY-qDcXYT8/edit#heading=h.1nre5zseobfj

PR #505, opened 2024-09-17
feat(consensus): re execute consensus when validators dont agree with the leader

AgustinRamiroDiaz:
Fixes https://github.com/yeagerai/genlayer-simulator/issues/168

# What

- leader gets rotated when majority of votes is not achieved
- transactions end in `UNDETERMINDED` if consensus is not reached
- refactors

# Why

to handle a feature of the consensus algorithm that is part of the spec we need to simulate

# Testing done

- Added unit tests

# Decisions made

- Created `Transaction` type in order to improve developer experience. 
  - I didn't made all the changes everywhere to not make a huge refactor
- Added all logic in the `exec_transaction`
  - I think that it's easier to read than having multiple functions with many moving parameters
  - Probably at some point this would be it's own class, written like a state machine to decouple all transitions into more manageable chunks. I think that refactor is out of scope ATM 
- Also migrated `value` from `float` to `int`

# Checks

- [x] I have tested this code

# Reviewing tips

- Don't check commits
- Read tests carefully

# User facing release notes

The consensus algorithm now re-executes when majority of votes is not reached. This is done by rotating the leader among the validators. This is an improvement towards simulating closer the actual behavior of GenLayer


backend/consensus/base.py
```diff
@@ -87,14 +123,24 @@ async def _run_consensus(self):
...
+        node_factory: Callable[
+            [
+                dict,
+                ExecutionMode,
+                ContractSnapshot,
+                Receipt | None,
+                MessageHandler,
+            ],
+            Node,
+        ] = node_factory,
```
AgustinRamiroDiaz, line 140: changes for dependency injection in tests

backend/consensus/base.py
```diff
@@ -104,118 +150,113 @@ async def exec_transaction(
...
+                len([vote for vote in votes.values() if vote == Vote.AGREE.value])
+                >= num_validators // 2
+            ):
+                break  # Consensus reached
+
+            print(
+                "Consensus not reached for transaction, rotating leader: ", transaction
+            )
+
+        else:  # this block is executed if the loop above is not broken
```
AgustinRamiroDiaz, line 234: the `for else` might be confusing at first, but I think it's exactly what we need here. I'm open to alternatives

backend/protocol_rpc/requirements.txt
```diff
@@ -8,7 +8,8 @@ python-dotenv==1.0.1
 asgiref==3.7.2
 pydantic==2.6.4
 numpy==1.26.4
-pytest==8.1.1
+pytest==8.3.3
+pytest-asyncio==0.24.0
```
AgustinRamiroDiaz, line 12: now we can unit test async functions

backend/consensus/base.py
```diff
@@ -76,7 +110,9 @@ async def _run_consensus(self):
                                     TransactionsProcessor(session),
                                     ChainSnapshot(session),
                                     AccountsManager(session),
-                                    session,
+                                    lambda contract_address, session=session: ContractSnapshot(
```
cristiam86, line 113: what is this magic? ^^

tests/unit/test_vrf.py
```diff
@@ -0,0 +1,46 @@
+from backend.consensus.vrf import get_validators_for_transaction
+
+
+def list_of_dicst_to_set(list_of_dicts: list[dict]) -> set:
```
cristiam86, line 4: @AgustinRamiroDiaz I'm not sure about these randomized tests but I guess there is no other way for these files

backend/consensus/base.py
AgustinRamiroDiaz: This is a factory function to get contract snapshots. This is used for dependency injection of the ContractSnapshot, allowing us to create unit tests for it

I wanted to go with just
```py
lambda contract_address: ContractSnapshot(
    contract_address, session
),
```

But Pylance was telling me that
`Add a parameter to the parent lambda function and use variable "session" as its default value; The value of "session" might change at the next loop iteration. [+2 locations]sonarlint(python:S1515)`

tests/unit/test_vrf.py
AgustinRamiroDiaz: I've added another test without randomization
I think that both of them are valuable, these tests are "black box" and the new one is "white box"

cristiam86 (approved):

PR #513, opened 2024-09-18
changed renovatebot default branch to staging

YeagerAI-Bulat:
# What

changed renovatebot default branch to staging

# Why

Dependabot uses `default` branch, where it will create all PRs, which is `main`, but we use `staging` for merging all PRs


cristiam86 (changes requested):
We are removing staging and switch to trunk based development

PR #521, opened 2024-09-18
feat(GenVM): allow async calls in the contract constructor's __init__

### LINKED ISSUE #449: Allow async calls in the contract constructor
# Issue

This is a problem for contracts that want to call async functions in init methods, like LLM calls

this seems to be an issue related to coroutines: [the code instantiates the class](https://github.com/yeagerai/genlayer-simulator/blob/7f296f3eb6546fd05578174d174c131a31bb6f93/backend/node/genvm/base.py#L120) as is, but in order to actually call the llm you need to make init method async, which means that now the class' init needs to be awaited

# Some ideas
a) try and reuse the asyncio event loop to call class __init__ methods with asyncio
b) handle both async and sync __init__ in the simulator

In general people don't recommend using async inits, so maybe we need to expose the asyncio event loop in some way

### LINKED ISSUE #450: Allow having equivalence principles in the contract constructor

AgustinRamiroDiaz:
Fixes #450
Fixes #449

# What

- Added logic to handle `async def __init__` in contracts
- Copied logic from `run_contract` into `deploy_contract` to allow calling llms

# Why

This is a natural feature that developers may need in their contracts

# Testing done

- added e2e test

# Decisions made

- `deploy_contract` and `run_contract` could be refactored to share most of the behavior. I didn't wanted to do it to not make it more complex for reading, and we'll also be removing that piece of code in the future
- fixed 2 issues in one PR since they are highly related

# Checks

- [x] I have tested this code

# Reviewing tips

learn about the `__new__` function in python https://stackoverflow.com/questions/8106900/new-and-init-in-python

# User facing release notes

You can now have `async def __init__` methods in your contracts, which empowers you to call LLMs at deploy time :rocket:


.github/workflows/backend_integration_tests_pr.yml
```diff
@@ -12,7 +12,7 @@ on:
       - edited
 
 concurrency:
-  group: backend-e2e-${{ github.event.number }}
+  group: backend-e2e-${{ github.event.number || github.event.pull_request.number}}
```
AgustinRamiroDiaz, line 15: extra: I saw that sometimes these tests were getting canceled. I suspect is due to `github.event.number` being null on some cases

tests/integration/conftest.py
```diff
@@ -17,3 +20,9 @@ def setup_validators():
         payload("delete_all_validators")
     ).json()
     assert has_success_status(delete_validators_result)
+
+
+@pytest.fixture
+def from_account() -> Iterator[Account]:
```
cristiam86, line 26: @AgustinRamiroDiaz could you please clarify on this?

cristiam86 (approved):

cristiam86 (approved):

denishacquin (approved):

PR #540, opened 2024-09-25
feat: improve logging system in terminal and UI

### LINKED ISSUE #436: Return validator execution logs

### LINKED ISSUE #448: Show captured outputs also in the dev terminal
Right now, the execution logs are getting captured and forwarded through the websocket, but we also need to continue printing those in the system terminal

denishacquin:
Fixes #448 
Fixes #436 

# What

- Refactored and improved the overall logging system for RPC, GenVM, consensus, both in backend and in frontend.
- Display live transaction status updates in both terminal and frontend logs
- Added filtering capabilities in the UI
- Added ability to filter by tx hash directly from transaction items
- Removed some unused logic in the frontend

Here's the list of events that trigger logs:

```
const trackedEvents = [
    'endpoint_call',
    'endpoint_success',
    'endpoint_error',
    'transaction_status_updated',
    'consensus_data_updated',
    'read_contract',
    'write_contract',
    'write_contract_failed',
    'deploying_contract',
    'deployed_contract',
    'contract_deployment_failed',
  ];
```

image
image

# Why

- To improve debugging and overall developer experience

# Testing done

- e2e
- unit
- integration
- error handling for each case

# Decisions made

- Added **loguru** python package for easy colored logs
- Introduced some types for the new events in the backend to enforce type safety
- Decided to keep **llm calls** and **get page** for later as the context is different more tricky to implement

# Checks

- [x] I have tested this code

# Reviewing tips

- Need special attention on the message handler class
- Check out the terminal and play around with the logs to try the new features
- Not 100% sure we should print the payloads in the terminal, it's a lot of clutter (we were doing it before)

# User facing release notes

- Improved logging of key events in the backend terminal
- Improved logging in the UI
- Added filtering capabilities in the logs in the UI


denishacquin:
Fixed conflicts

backend/node/genvm/base.py
```diff
@@ -222,24 +235,45 @@ async def run_contract(
                 else:
                     function_to_run(*args)
             except Exception as e:
-                print("\nError running contract", e)
-                print(f"\n{traceback.format_exc()}")
+                trace = traceback.format_exc()
                 error = e
+                print("Error executing method", error)
```
cristiam86, line 240: @denishacquin where are these logs getting printed?

backend/database_handler/transactions_processor.py
```diff
@@ -149,9 +161,24 @@ def set_transaction_result(self, transaction_hash: str, consensus_data: dict):
...
-            "Updating transaction status",
-            transaction_hash,
-            TransactionStatus.FINALIZED.value,
-        )
         self.session.commit()
+        self.dispatch_transaction_status_update(
+            transaction_hash, TransactionStatus.FINALIZED
+        )
+
+    def dispatch_transaction_status_update(
```
cristiam86, line 169: @denishacquin I'm not 100% sure that this should be here. I think it would be better not to have the msg_handler injected into the transactions_processor and log the transaction transitions in the consensus. 
What do you think @AgustinRamiroDiaz?

cristiam86:
@denishacquin could you please explain if the logs are getting printed in the system terminal despite capturing the std_output because we are using loguru?

backend/database_handler/transactions_processor.py
denishacquin: makes sense, moved it to consensus and refactored the actual tx update together with it to reduce repetition but might be overkill - let me know what you think

backend/node/genvm/base.py
denishacquin: In the terminal. The intention here was merely to print the error and trace directly to the terminal when it happens but it might be redundant with the "pretty" that happen below.

denishacquin: Nevermind, they actually only get sent back with the send_message.

How about we print the stdout directly to the terminal if it exists (for clarity of debugging backend via the terminal) and keep sending to the log event bus like so:
image

Resulting in this in terminal (ok vs error):
image
Screenshot 2024-09-27 at 12 23 42

And this in frontend (ok vs error):
image
image

Also added a check to only print the "OK" message when execution is successful.

denishacquin:
> @denishacquin could you please explain if the logs are getting printed in the system terminal despite capturing the std_output because we are using loguru?

Not 100% I understand here, let's 

> @denishacquin could you please explain if the logs are getting printed in the system terminal despite capturing the std_output because we are using loguru?

please check comments @cristiam86 - if not clear, let's discuss it! :)

backend/node/genvm/base.py
cristiam86: OK I think that is ok to also print the errors in the terminal for a better debugging

denishacquin:
> @denishacquin could you please explain if the logs are getting printed in the system terminal despite capturing the std_output because we are using loguru?

image

cristiam86 (approved):
Good Job, you just need to solve the conflicts.
The multiline logs are now printed like different log records, but this is fine for now.
Screenshot 2024-09-30 at 00 18 09

denishacquin:
> Good Job, you just need to solve the conflicts. The multiline logs are now printed like different log records, but this is fine for now. Screenshot 2024-09-30 at 00 18 09

Fixed the multi line logs like so:
image

cristiam86 (approved):

backend/protocol_rpc/message_handler/base.py
```diff
@@ -1,65 +1,66 @@
...
-            function_result = (
-                (result_string[:MAX_LOG_MESSAGE_LENGTH] + "...")
-                if result_string is not None
-                and len(result_string) > MAX_LOG_MESSAGE_LENGTH
-                else result_string
-            )
-            log_method(function_name + ": " + function_result)
-        else:
-            raise Exception(f"Logger does not have the method {result.status}")
+    def socket_emit(self, log_event: LogEvent):
```
AgustinRamiroDiaz, line 27: nit: prepending with `_` is a way to mark internal methods in Python. It might be a good idea to do that so that we don't accidentally manually call this method for example

backend/protocol_rpc/message_handler/base.py
```diff
@@ -1,65 +1,66 @@
...
+        if log_event.data:
+            try:
+                data_str = json.dumps(log_event.data, default=lambda o: o.__dict__)
+                log_message += f" {gray}{data_str}{reset}"
+            except TypeError as e:
+                log_message += (
+                    f" {gray}{str(log_event.data)} (serialization error: {e}){reset}"
+                )
+
+        log_method(log_message)
```
AgustinRamiroDiaz, line 58: nit: this method is hard to understand for me, and I feel it's like reinventing the wheel at some point. Doesn't `loguru` have formatting features for logs?

backend/protocol_rpc/message_handler/base.py
```diff
@@ -1,65 +1,66 @@
...
+
+    def send_message(self, log_event: LogEvent, log_to_terminal: bool = True):
+        if log_to_terminal:
+            self.log_message(log_event)
+        self.socket_emit(log_event)
```
AgustinRamiroDiaz, line 63: nit: maybe we can use loguru sinks to handle multiple outputs https://loguru.readthedocs.io/en/stable/api/logger.html#sink

backend/consensus/base.py
```diff
@@ -352,10 +390,32 @@ def execute_transfer(
...
+        self.dispatch_transaction_status_update(
+            transactions_processor, transaction.hash, TransactionStatus.FINALIZED
         )
 
+    def dispatch_transaction_status_update(
+        self,
+        transactions_processor: TransactionsProcessor,
+        transaction_hash: str,
+        new_status: TransactionStatus,
+    ):
```
AgustinRamiroDiaz, line 402: nit: I know this was done to have DRY code, but I feel that creating a function just to log this might be an anti pattern. Maybe it makes more sense to have this log inside `transactions_processor.update_transaction_status`. WDYT?

backend/node/genvm/base.py
```diff
@@ -170,33 +176,47 @@ async def deploy_contract(
                 )
 
             except Exception as e:
-                print("Error deploying contract", e)
                 trace = traceback.format_exc()
-                print(trace)
                 error = e
```
AgustinRamiroDiaz, line 180:
```suggestion
            except Exception as error:
                trace = traceback.format_exc()
```

backend/node/genvm/base.py
```diff
@@ -258,24 +279,51 @@ async def run_contract(
                 else:
                     function_to_run(*args)
             except Exception as e:
-                print("\nError running contract", e)
-                print(f"\n{traceback.format_exc()}")
+                trace = traceback.format_exc()
                 error = e
```
AgustinRamiroDiaz, line 283:
```suggestion
            except Exception as error:
                trace = traceback.format_exc()
```

backend/node/genvm/base.py
```diff
@@ -258,24 +279,51 @@ async def run_contract(
...
+                print("Error executing method", error)
+                print(trace)
```
AgustinRamiroDiaz, line 285: if we have the logger below, I don't think we should have prints. I think we should go only with the logger

backend/protocol_rpc/message_handler/types.py
```diff
@@ -1,17 +1,33 @@
+from enum import Enum
 from dataclasses import dataclass
 
-from backend.protocol_rpc.types import EndpointResult
+
+class EventType(Enum):
+    DEBUG = "debug"
+    INFO = "info"
+    SUCCESS = "success"
+    ERROR = "error"
```
AgustinRamiroDiaz, line 9: shouldn't we use event types from `loguru` or `logging` packages?

AgustinRamiroDiaz (approved):
Awesome work! It looks :100:

All my comments are minor, the behavior works great

AgustinRamiroDiaz:
@denishacquin please also fix backend unit tests

backend/protocol_rpc/message_handler/base.py
denishacquin: ok, I shall do the same for both methods then

denishacquin: image

backend/consensus/base.py
denishacquin: I felt the same way when doing it, I think you're right for the anti pattern. Originally I was doing it in the processor but @cristiam86 mentioned this should be in consensus. So we have anti pattern VS separation of concern.
We have two other options:
- Do like you said and like I did before @AgustinRamiroDiaz and move it to the transactions_processor.update_transaction_status method (no repeated logic, just a little bit of weird concern separation since we log a consensus event from the processor) if that's OK with @cristiam86 
- Repeat the log event at every place in the code where we update a tx status (repeated logic, but no anti pattern and keep concern in consensus)

AgustinRamiroDiaz: Let's leave it like this for now since
- we don't have a clear agreement
- it's a nit
- we need to go fast for the hackathon

Thanks for taking this into consideration :heart:

backend/node/genvm/base.py
cristiam86: We want this to be in the developer terminal, why are you suggesting to remove it?

backend/protocol_rpc/message_handler/base.py
denishacquin: I though about it but unfortunately a sink only takes a Message, meaning I would lose the structuring of the data (LogEvent) and thus have to re-extract the data from the message somehow.

backend/protocol_rpc/message_handler/types.py
denishacquin: loguru doesn't have actual types that I can import it seems, rather they have pre-defined and customizable levels but I don't think I can simply import the type without writing some logic - happy to be wrong tho

backend/node/genvm/base.py
denishacquin: This is a little confusing I admit, but basically those get buffered here and a little down they get printed directly to terminal indeed. The reasoning is to avoid fancy formatting for better readability of multiine tracebacks in this instance, and in parallel also sentd to the frontend of course.

backend/protocol_rpc/message_handler/base.py
denishacquin: I feel you. This logic was here before and I actually tried to remove it once I implemented loguru, but I'm hitting some limitations:
- Loguru doesn't have a truncation safety net, so we need to add that in a custom way I guess;
- I can't use loguru's color markup in the message, only in the .add() declaration, which does not help with what I'm trying to achieve here (having different coloring (white) for the message and (gray) for the data so as to improve readability in the terminal)

Maybe there are ways to improve this, but I haven't found how yet. We could also potentially just remove the printing of the data altogether and only keep the basic message and not print all the payload data.

backend/node/genvm/base.py
denishacquin: actually need the remapping of `error`  here because it is used down below when returning the receipt

backend/node/genvm/base.py
denishacquin: same- actually need the remapping of `error`  here because it is used down below when returning the receipt

backend/database_handler/transactions_processor.py
```diff
@@ -149,7 +149,6 @@ def set_transaction_result(self, transaction_hash: str, consensus_data: dict):
             self.session.query(Transactions).filter_by(hash=transaction_hash).one()
         )
 
-        transaction.status = TransactionStatus.FINALIZED
```
AgustinRamiroDiaz, line 149: What's the reason behind this change?

denishacquin: It was redundant because we also update the status in the update status method, and I thought it was causing the unit test to fail. But actually updating the mocking in the test was enough, so I'm reverting this! nevermind

AgustinRamiroDiaz (approved):

backend/database_handler/migration/versions/a32f85df2806_add_client_session_id_to_transactions.py
```diff
@@ -0,0 +1,34 @@
...
+down_revision: Union[str, None] = "3566595124f6"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "transactions",
+        sa.Column("client_session_id", sa.String(length=255), nullable=True),
```
cristiam86, line 26: @AgustinRamiroDiaz @denishacquin This works, but it doesn't comply with the transactions and the consensus we are simulating here. 
I would say that for now, it is ok if it works as expected, but we need to search for a solution at the WebSockets connection level instead of changing the transactions model.

backend/node/base.py
```diff
@@ -89,27 +89,14 @@ async def run_contract(
...
+        result = self.genvm.get_contract_data(
             code,
             state,
             method_name,
             method_args,
             self.contract_snapshot_factory,
-            output_buffer,
         )
 
-        if self.genvm.contract_runner.mode == ExecutionMode.LEADER:
```
cristiam86, line 103: @AgustinRamiroDiaz @denishacquin could you please point to the place where this is being done now?

backend/consensus/base.py
```diff
@@ -333,8 +392,10 @@ def execute_transfer(
 
             # If the sender does not have enough balance, set the transaction status to UNDETERMINED
             if from_balance < transaction.value:
-                transactions_processor.update_transaction_status(
-                    transaction.hash, TransactionStatus.UNDETERMINED
+                MessageHandler.dispatch_transaction_status_update(
```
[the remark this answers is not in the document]
AgustinRamiroDiaz: you are right! will fix

backend/database_handler/migration/versions/a32f85df2806_add_client_session_id_to_transactions.py
AgustinRamiroDiaz: Fully agree.
This was the quickest solution. I think that the long term solution would be that the frontend should subscribe to the transactions that it wants to see, and the backend should broadcast to all subscribers

backend/node/base.py
AgustinRamiroDiaz: It was the fastest way at the moment, but I think that we'll eventually move all logging outside of the GenVM

cristiam86 (approved):

cristiam86 (approved):

PR #544, opened 2024-09-26
fix: made optional the from param in call contract method

### LINKED ISSUE #543: Make optional the from param in call contract method

cristiam86:
Fixes #543 

# What

- from param in call method is optional

# Why

- GenLayerJS may not send this param

# Testing done

- tested the bug fix

# Decisions made

# Checks

- [X] I have tested this code
- [X] I have reviewed my own PR
- [X] I have created an issue for this PR
- [X] I have set a descriptive PR title compliant with [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

# Reviewing tips

# User facing release notes

Fixed an issue when not sending the from param in the call contract method


AgustinRamiroDiaz (commented):
Changes look good, but why do we have it in the first place if it can be nulled? Maybe we should remove it altogether, or at least remove the check since it's bypassable

AgustinRamiroDiaz (approved):

cristiam86:
> Changes look good, but why do we have it in the first place if it can be nulled? Maybe we should remove it altogether, or at least remove the check since it's bypassable

I'm not sure if I follow. The param must be optional since it can be sent from GenLayerJS. Then, if an address is passed, we need to check that it is valid

AgustinRamiroDiaz (approved):

Issue #454, opened 2024-09-02
Contract-to-contract write


cristiam86:
Allow to send a transaction from one contract to another contract

AgustinRamiroDiaz:
Based on https://docs.google.com/document/d/1qUkQpR3urVNXJvlucmEfeOQJH510j7_XOqY-qDcXYT8/edit#heading=h.jdxv42hmj9qs

We should be able to call something like

```
Contract("0x123").send_message(on_finalized_or_on_accepted, value, "func_to_call", **args)
```
---

Node Receipts should return two arrays containing the on_finalized and on_accepted pending transactions, which will be processed in the consensus `exec_transaction` when the transaction reaches those states

PR #552, opened 2024-10-01
chore: improve contract to contract write

AgustinRamiroDiaz:
Extension of #454

# What

Adds self relationship to `Transactions` to be able to track the triggered transactions
I'll try adding more tests through the day if I have time

# Why

- to improve tracability

# Testing done

- improved e2e tests
- manually tested

# Decisions made

In the UI it's just plain text

image

We could improve this maybe by polling the `triggered_transactions`. But those transactions could trigger even more transactions, so this needs a bit more of definition prior to working on it

# Checks

- [x] I have tested this code

# Reviewing tips

- https://docs.sqlalchemy.org/en/20/orm/self_referential.html
- https://docs.sqlalchemy.org/en/20/_modules/examples/adjacency_list/adjacency_list.html

# User facing release notes

You can now see the transactions triggered by your Intelligen Contract when writing to other contracts


backend/database_handler/transactions_processor.py
```diff
@@ -103,6 +111,11 @@ def insert_transaction(
             s=None,
             v=None,
             leader_only=leader_only,
+            triggered_by=(
+                self.session.query(Transactions).filter_by(hash=triggered_by_hash).one()
+                if triggered_by_hash
+                else None
+            ),
```
AgustinRamiroDiaz, line 118: This query is necessary since SQLAlchemy works with objects, therefore we need to have the object related here

cristiam86 (approved):

denishacquin (approved):

backend/database_handler/transactions_processor.py
```diff
@@ -155,4 +167,8 @@ def set_transaction_result(self, transaction_hash: str, consensus_data: dict):
         transaction.status = TransactionStatus.FINALIZED
         transaction.consensus_data = consensus_data
 
-        self.session.commit()
+        print(
```
cristiam86, line 170: @AgustinRamiroDiaz wasn't this moved to the consensus? and why is there a bare print here?

backend/database_handler/transactions_processor.py
```diff
@@ -35,6 +35,11 @@ def _parse_transaction_data(transaction_data: Transactions) -> dict:
             "created_at": transaction_data.created_at.isoformat(),
             "leader_only": transaction_data.leader_only,
             "client_session_id": transaction_data.client_session_id,
+            "triggered_by": transaction_data.triggered_by_hash,
+            "triggered_transactions": [
```
cristiam86, line 39: @AgustinRamiroDiaz when is this field populated?

PR #554, opened 2024-10-02
feat: add nonces to transactions

denishacquin:
Fixes #issue-number-here

# What

- changed thing a for b
- also did this other unrelated thing in my path

# Why

- to fix a bug

# Testing done

- tested the new feature
- tested the bug fix

# Decisions made

# Checks

- [ ] I have tested this code

# Reviewing tips

# User facing release notes


frontend/src/components/Simulator/TransactionsList.vue
```diff
@@ -38,7 +43,10 @@ const handleClearTransactions = () => {
 
 <template>
   <PageSection data-testid="latest-transactions">
-    <template #title>Transactions</template>
+    <template #title
+      >Transactions ({{ accountsStore.transactionCount }})</template
```
AgustinRamiroDiaz, line 47: !!! This is not rendering properly. Please fix before merging

frontend/src/hooks/useWebSocketClient.ts
```diff
@@ -1,14 +1,18 @@
...
+import { io, type Socket } from 'socket.io-client';
 
-const webSocketClient = io(import.meta.env.VITE_WS_SERVER_URL);
+let webSocketClient: Socket | null = null;
 
 export function useWebSocketClient() {
+  if (!webSocketClient) {
+    webSocketClient = io(import.meta.env.VITE_WS_SERVER_URL);
+  }
+
```
AgustinRamiroDiaz, line 9: curiosity: what's this change for?

AgustinRamiroDiaz (approved):

backend/database_handler/transactions_processor.py
```diff
@@ -78,14 +80,19 @@ def insert_transaction(
...
-        )
+        # Only allow transactions without from_address for 'fund_account'
+        if not from_address and type is not TransactionType.SEND.value:
+            raise Exception("From address is required")
+
+        if from_address:
+            current_nonce = self.get_transaction_count(from_address)
+
+            if nonce != current_nonce:
+                raise Exception(f"Nonce is not current nonce. Provided: {nonce}")
```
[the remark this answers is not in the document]
denishacquin: had it before, good catch, done!

frontend/src/hooks/useWebSocketClient.ts
denishacquin: This is introducing a singleton pattern that allowed to make sure we only initiate the client once, and also allowed to fix the related broken unit test at the same time.

frontend/src/components/Simulator/TransactionsList.vue
denishacquin: this was for testing purpose and forgot to remove it after I deleted the variable - great catch, thanks!

AgustinRamiroDiaz (approved):

denishacquin:
Kindly requesting your review as well @cristiam86 to validate the functionality

cristiam86:
> Kindly requesting your review as well @cristiam86 to validate the functionality

sure, I'm on it

backend/database_handler/transactions_processor.py
```diff
@@ -78,14 +80,21 @@ def insert_transaction(
...
+        nonce: int,
         leader_only: bool,
         client_session_id: str | None,
     ) -> int:
-        nonce = (
-            self.session.query(Transactions)
-            .filter(Transactions.from_address == from_address)
-            .count()
-        )
+        # Only allow transactions without from_address for 'fund_account'
```
cristiam86, line 87: @denishacquin this type of validation would be better in the JSON RPC endpoint. It is the place where we are adding this types of validations so far.

backend/database_handler/transactions_processor.py
```diff
@@ -78,14 +80,21 @@ def insert_transaction(
...
+        if not from_address and type is not TransactionType.SEND.value:
+            raise Exception("From address is required")
+
+        if from_address:
+            current_nonce = self.get_transaction_count(from_address)
+
+            if nonce != current_nonce:
+                raise Exception(
```
cristiam86, line 95: @denishacquin This one seems ok because it is a business logic validation

frontend/src/components/Simulator/TransactionsList.vue
```diff
@@ -1,12 +1,17 @@
...
+  useContractsStore,
+  useTransactionsStore,
+} from '@/stores';
 import { TrashIcon } from '@heroicons/vue/24/solid';
 import TransactionItem from './TransactionItem.vue';
 import PageSection from './PageSection.vue';
 import EmptyListPlaceholder from '@/components/Simulator/EmptyListPlaceholder.vue';
 
 const contractsStore = useContractsStore();
+const accountsStore = useAccountsStore();
```
cristiam86, line 14: @denishacquin why is this used for?

cristiam86 (changes requested):

frontend/src/components/Simulator/TransactionsList.vue
denishacquin: good spot, removing

backend/database_handler/transactions_processor.py
denishacquin: You're right - however, I was able to simply remove this check by passing the nonce explicitly from the fund_account endpoint method.

Note that it's weird to get a nonce from an empty address, but I guess this is just a hacky case we need to support for now to allow funding accounts, as later on we will probably be using faucets.

cristiam86 (approved):

PR #558, opened 2024-10-07
feat: migrate to new calldata

kp2pml30:
Fixes \<no issue\>

# What

- implemented new calldata encoding

# Why

- first step for migration to new genvm

# Testing done

- python file is taken from genvm, untested here
- ts parser is tested to some extent
- some manual testing was done

# Decisions made

# Checks

- [x] I have tested this code

# Reviewing tips

For now there is a problem with "json" input by user for obvious reasons. I can allocate some time to fix them, but I am not great in reactive ui's

# User facing release notes

New calldata string representation is slightly different from json:
- it has bytes type (encoded in text as `b#`\<hex bytes>)
- it has address type (encoded in text as `addr#`\<64 hex digits>
- keys can omit `"`


frontend/src/calldata/encoder.ts
```diff
@@ -0,0 +1,161 @@
...
+
+function encodeMap(to: number[], arr: Iterable<[string, CalldataEncodable]>) {
+  const newEntries = Array.from(
+    arr,
+    ([k, v]): [Uint8Array, CalldataEncodable] => [
+      new TextEncoder().encode(k),
+      v,
+    ],
+  );
+  newEntries.sort(([k1, _v1], [k2, _v2]) => compareString(k1, k2));
```
kp2pml30, line 70: eslint gives a warning about `_v1` and `_v2` being unused even though they are prefixed with an underscore. How to silence that?...

backend/database_handler/transactions_processor.py
```diff
@@ -103,7 +114,7 @@ def insert_transaction(
             hash=transaction_hash,
             from_address=from_address,
             to_address=to_address,
-            data=data,
+            data=json.loads(self._transaction_data_to_str(data)),
```
kp2pml30, line 117: Explanation: this converts bytes to base64

tests/common/request.py
```diff
@@ -75,16 +78,19 @@ def send_transaction(
...
 
 
 def deploy_intelligent_contract(
-    account: Account, contract_code: str, constructor_params: str
+    account: Account, contract_code: str, method_args: list
 ) -> tuple[str, dict]:
-    deploy_data = [contract_code, constructor_params]
+    deploy_data = [
+        contract_code,
+        calldata.encode({"method": "__init__", "args": method_args}),
```
kp2pml30, line 92: For now I would prefer to stick to hardcoded `__init__` for a constructor. I am thinking about changing some abi properties of calldata, so we may change it later

frontend/src/calldata/parser.ts
```diff
@@ -0,0 +1,211 @@
...
+  [true, /^b#[a-fA-F0-9]*/g, TokenKind.Bytes],
+  [true, /^\{/g, TokenKind.LCpar],
+  [true, /^\[/g, TokenKind.LSpar],
+  [true, /^\}/g, TokenKind.RCpar],
+  [true, /^\]/g, TokenKind.RSpar],
+  [true, /^,/g, TokenKind.Comma],
+  [true, /^:/g, TokenKind.Colon],
+  [true, /^\w[\w0-9]*/g, TokenKind.Id],
+  [true, /^'([^']|\\.)*'/g, TokenKind.Str],
+  [true, /^"([^"]|\\.)*"/g, TokenKind.Str],
```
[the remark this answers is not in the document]
kp2pml30: should be fixed

backend/node/genvm/base.py
```diff
@@ -492,8 +511,9 @@ def get_contract_data(
                         EventScope.GENVM,
                         "Call method: " + method_name,
                         {
-                            "method_name": method_name,
-                            "method_args": method_args,
+                            "calldata": str(
```
cristiam86, line 514: @kp2pml30 this is for logging purposes in the Simulator's frontend's UI. Can't we send the method name and args as before? is it because of their serialization?

kp2pml30: Hmmm. There is an issue that you can't `json.dumps` them, but in general it is possible to convert it to a string (with both f string or writing that serialization)

frontend/src/calldata/encoder.ts
cristiam86: @kp2pml30 they are not being used, when calling compareString(k1, k2) so you can totally omit them.

kp2pml30: Okay, js pattern matching is as surprising as always

cristiam86 (commented):
Awesome @kp2pml30 do you think that it would be possible to add some comments to frontend/src/calldata/encoder.ts and frontend/src/calldata/parser.ts so we can understand it better?

backend/node/genvm/base.py
kp2pml30: Implemented string representation in python and now it is dumped

Example:
```
genlayer-simulator-jsonrpc-1             | SUCCESS  | [GenVM] Deploying contract {"calldata": "{'args':[],'method':'__init__',}", "output": "Hello! How can I assist you today?\n"}
```

backend/database_handler/transactions_processor.py
```diff
@@ -42,6 +43,15 @@ def _parse_transaction_data(transaction_data: Transactions) -> dict:
             ],
         }
 
+    @staticmethod
+    def _transaction_data_to_str(data: dict) -> str:
```
AgustinRamiroDiaz, line 47: Awesome! Could you add some function docstring explaining a bit why are we serializing bytes as base64?

backend/node/genvm/base.py
```diff
@@ -47,6 +53,15 @@ def safe_globals(override_globals: dict[str] = None):
         globals().update(old_globals)
 
 
+_FAKE_DECODED_DATA = object()
```
AgustinRamiroDiaz, line 56: what's this for? 

seems like you may want to check if some variables have or not been initialized, I think using `None` is a better approach

backend/node/genvm/calldata.py
```diff
@@ -0,0 +1,198 @@
+from .types import Address
```
AgustinRamiroDiaz, line 1: I'm assuming this file is copied/adapted from the GenVM codebase right? It's very cryptic to me

backend/node/genvm/types.py
```diff
@@ -21,15 +55,19 @@ class ExecutionResultStatus(Enum):
...
 class PendingTransaction:
     address: str  # Address of the contract to call
-    method_name: str
-    args: list
+    calldata: bytes
+
+    def to_dict(self):
+        return {
+            "address": self.address,
+            "calldata": str(base64.b64encode(self.calldata), encoding="ascii"),
```
AgustinRamiroDiaz, line 63: maybe it's a good idea for code organiztion to have a function in the GenVM module called something like "encode_calldata" so that we don't repeat always `b64encode...` and we have one source of truth to integrate with the GenVM

frontend/src/calldata/encoder.ts
```diff
@@ -0,0 +1,164 @@
+import type { CalldataEncodable } from './types';
```
AgustinRamiroDiaz, line 1: why is this whole module needed in the frontend? will it be part of GenLayerJS?

AgustinRamiroDiaz (commented):
Changes in general look good. I'm super lost around the encoding part

I have checked this branch and I cannot use the simulator properly, when trying to run write methods I get

```
Uncaught (in promise) Error: Error writing to contract
    callWriteMethod useContractQueries.ts:220
    handleCallWriteMethod ContractMethodItem.vue:58
    callWithErrorHandling runtime-core.esm-bundler.js:199
    callWithAsyncErrorHandling runtime-core.esm-bundler.js:206
    emit runtime-core.esm-bundler.js:6306
    createSetupContext/get emit/< runtime-core.esm-bundler.js:8016
    0 Btn.vue:46
    callWithErrorHandling runtime-core.esm-bundler.js:199
    callWithAsyncErrorHandling runtime-core.esm-bundler.js:206
    invoker runtime-dom.esm-bundler.js:722
```

I think the UI it's also broken in some places 
image




# Extra


I also want to remark that this PR has big breaking changes that will impact:
- GenLayerJS
- users upgrading

frontend/src/calldata/encoder.ts
cristiam86: yes, it will

cristiam86:
> Changes in general look good. I'm super lost around the encoding part
> 
> I have checked this branch and I cannot use the simulator properly, when trying to run write methods I get
> 
> ```
> Uncaught (in promise) Error: Error writing to contract
>     callWriteMethod useContractQueries.ts:220
>     handleCallWriteMethod ContractMethodItem.vue:58
>     callWithErrorHandling runtime-core.esm-bundler.js:199
>     callWithAsyncErrorHandling runtime-core.esm-bundler.js:206
>     emit runtime-core.esm-bundler.js:6306
>     createSetupContext/get emit/< runtime-core.esm-bundler.js:8016
>     0 Btn.vue:46
>     callWithErrorHandling runtime-core.esm-bundler.js:199
>     callWithAsyncErrorHandling runtime-core.esm-bundler.js:206
>     invoker runtime-dom.esm-bundler.js:722
> ```
> 
> I think the UI it's also broken in some places image
> 
> # Extra
> I also want to remark that this PR has big breaking changes that will impact:
> 
> * GenLayerJS
> * users upgrading

@AgustinRamiroDiaz Can you please disclose the contract and the method that you are calling to get this error? I've tried different ones and they all succeed.

backend/node/genvm/base.py
kp2pml30: I think no, because `None` can be encoded in calldata (even though it is invalid here). So I wish to distinguish between parsed calldata and unparsed

backend/database_handler/transactions_processor.py
kp2pml30: Because json.dumps doesn't support bytes?...

backend/node/genvm/calldata.py
kp2pml30: Yes, it is
Unfortunately genvm repo is still private, even though I asked JM about making it public. Docs for calldata are [there](https://github.com/yeagerai/genvm/blob/main/doc/calldata.md)

backend/node/genvm/types.py
kp2pml30: Considering that next pr will completely replace genvm folder and is more than half done... I am not sure that it is required right now

backend/database_handler/transactions_processor.py
AgustinRamiroDiaz: didn't know that! I think it's not common knowledge, so I'd love to have a comment there mentioning that :)

backend/node/genvm/base.py
AgustinRamiroDiaz: and can `object()` be encoded in calldata?

AgustinRamiroDiaz (approved):
The UI errors I mentioned turned out to be due to my local install, sorry for the confusion

Great work Kira!

backend/node/genvm/base.py
kp2pml30: This object is checked with `is`, so it is checked for a reference, and other code can't get the same reference. This trick is common in java
Answering your question, python object can't, js object can I think (map from a string to something)

kp2pml30:
Actually I found out that probably what we mean by address is 20 bytes and not 32 as in this PR. It is sort of confusing:
- account address is 20 bytes
- word size (in eth) is 32 bytes, slot addresses in eth are also 32 bytes

kp2pml30:
Sorry I messed up this PR.... It won't reopen

PR #559, opened 2024-10-07
feat: support both explicit and implicit vector ids in vector store

### LINKED ISSUE #467: SIM-Allow developers to set their own ID for VectorStore database items

denishacquin:
Fixes #467 

# What

- Allow explicitly setting vector IDs in the vector store
- Still support implicit / auto-incremental fallback

# Why

- to add more flexibility to the developers

# Testing done

- tested the different cases of explicit / implicit ID

# Decisions made

- Kept both explicit and implicit ID for flexibility

# Checks

- [x] I have tested this code

# Reviewing tips

- Try it out with the log_indexer example contract

# User facing release notes

- Vector store now also supports passing explicit vector ID when adding new vectors

### DIFF
--- a/backend/node/genvm/std/vector_store.py
+++ b/backend/node/genvm/std/vector_store.py
@@ -17,21 +17,24 @@ class VectorStore:
         self.vector_data = {}  # Dictionary to store vectors
         self.metadata = {}  # Dictionary to store metadata
         self.model_name = model_name
-        self.next_id = 0
 
-    def add_text(self, text: str, metadata: Any):
+    def add_text(self, vector_id: int, text: str, metadata: Any):
         """
         Add a new text to the store with its metadata.
 
         Args:
             text (str): The text to be added.
             metadata (Any): The metadata.
+            vector_id (int, optional): The ID for the vector. If not provided, a new ID will be generated.
         """
+        if not vector_id:
+            vector_id = max(self.vector_data.keys(), default=0) + 1
+            print(f"generated id: {vector_id}")
+        elif vector_id in self.vector_data:
+            raise ValueError(f"Vector ID {vector_id} already exists")
 
         model = get_model(self.model_name)
         embedding = model.encode([text])[0]
-        vector_id = self.next_id
-        self.next_id += 1
 
         self.texts[vector_id] = text
         self.vector_data[vector_id] = embedding


backend/node/genvm/std/vector_store.py
```diff
@@ -17,21 +17,24 @@ def __init__(self, model_name: str = None):
...
+    def add_text(self, text: str, metadata: Any, vector_id: Optional[int] = None):
         """
         Add a new text to the store with its metadata.
 
         Args:
             text (str): The text to be added.
             metadata (Any): The metadata.
+            vector_id (int, optional): The ID for the vector. If not provided, a new ID will be generated.
         """
+        if not vector_id and not isinstance(vector_id, int):
```
AgustinRamiroDiaz, line 30: I think we should fail if we were to get an invalid `vector_id` instead of generating one manually, since it could be confusing for the user
Example: I try to create one with an id of `"123"`, but I didn't realize that I passed a string. So then I see that my vector has id `1`

frontend/src/components/Simulator/ContractMethodItem.vue
```diff
@@ -79,7 +79,7 @@ const resetInputs = () => {
     switch (input.type) {
       case 'uint256':
       case 'float':
-        defaultValue = 0;
+        defaultValue = '';
```
AgustinRamiroDiaz, line 82: This seems weird. Why this change?

denishacquin: Basically I wanted to allow passing an empty value to be able to test optional parameter. Does that make sense?

backend/node/genvm/std/vector_store.py
denishacquin: Good point, will adapt

denishacquin: changed to only allow integers

frontend/src/components/Simulator/ContractMethodItem.vue
denishacquin: I actually think this change was not necessary, reverting

cristiam86 (approved):

denishacquin:
@AgustinRamiroDiaz I adapted for your comments - pinging your for a quick final check if you don't mind :)

AgustinRamiroDiaz (approved):

PR #564, opened 2024-10-15
feat(deployment): add healthcheck for backend

AgustinRamiroDiaz:
# What

Adds a healthcheck for the backend service

# Why

- In other applications I want to build depending on the simulator (feedbuzz, intelligent oracles) I need to depend on this service in order to be able to make calls to it. Healthchecks are the way to go for specifying dependencies between containers
- Healthchecks are a great idea in general

# Testing done

- tested the new feature
- tested the bug fix

# Decisions made

- I've done it without using the `tests` module to not introduce module dependencies
- I've done it with a Python script since CURL nor WGET are available in the image

# Checks

- [x] I have tested this code

# Reviewing tips

none

# User facing release notes

The backend container now has a healthcheck! You can use this to depend on it in your applications building on top of GenLayer's simulator

### DIFF
--- a/backend/healthcheck.py
+++ b/backend/healthcheck.py
@@ -1,0 +1,34 @@
+"""
+File used to check the health of the backend API.
+We only check that the API is responding, and that the response is valid.
+"""
+
+import json
+import requests
+import argparse
+
+parser = argparse.ArgumentParser(description="Check the health of the backend API.")
+parser.add_argument(
+    "--port",
+    type=str,
+    nargs="?",
+    help="The port number for the API",
+)
+
+args = parser.parse_args()
+
+response = requests.post(
+    f"http://localhost:{args.port}/api",
+    data=json.dumps(
+        {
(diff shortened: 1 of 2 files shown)


cristiam86 (approved):

YeagerAI-Bulat:
I wanted to suggest adding a `/metrics` API endpoint that can expose metrics in Prometheus format. This would be really helpful for monitoring and would allow us to collect and visualize various metrics about the application's behavior.
In addition, we could use this `/metrics` endpoint to perform health checks. This approach could give us more detailed insights about system health compared to a simple "heartbeat" check, as it would expose more granular metrics, making it easier to track performance and identify issues early on.

AgustinRamiroDiaz:
@YeagerAI-Bulat I agree, but having a `/metrics` endpoint is a whole issue in of itself, given that we currently don't have any metrics

Could you create an issue to track this request of having Prometheus metrics?

PR #565, opened 2024-10-15
feat: migrate to new calldata

kp2pml30:
Fixes \<no issue\>

# What

- implemented new calldata encoding

# Why

- first step for migration to new genvm

# Testing done

- python file is taken from genvm, untested here
- ts parser is tested to some extent
- some manual testing was done

# Decisions made

# Checks

- [x] I have tested this code

# Reviewing tips

For now there is a problem with "json" input by user for obvious reasons. I can allocate some time to fix them, but I am not great in reactive ui's

# User facing release notes

New calldata string representation is slightly different from json:
- it has bytes type (encoded in text as `b#`\<hex bytes>)
- it has address type (encoded in text as `addr#`\<64 hex digits>
- keys can omit `"`

### DIFF
--- a/backend/database_handler/transactions_processor.py
+++ b/backend/database_handler/transactions_processor.py
@@ -7,6 +7,7 @@ from sqlalchemy.orm import Session
 from .models import TransactionStatus
 from eth_utils import to_bytes, keccak, is_address
 import json
+import base64
 
 
 class TransactionsProcessor:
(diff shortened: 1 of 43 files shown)


backend/database_handler/transactions_processor.py
```diff
@@ -42,6 +43,24 @@ def _parse_transaction_data(transaction_data: Transactions) -> dict:
             ],
         }
 
+    @staticmethod
+    def _transaction_data_to_str(data: dict) -> str:
+        """
```
kp2pml30, line 48: @AgustinRamiroDiaz comment added

kp2pml30:
Actually I found out that probably what we mean by address is 20 bytes and not 32 as in this PR. It is sort of confusing:
- account address is 20 bytes
- word size (in eth) is 32 bytes, slot addresses in eth are also 32 bytes

If I fix that right now, will it break js sdk? @cristiam86

PR #569, opened 2024-10-15
feat(rpc): add sim_getTransactionsForAddress to be able to get all transactions for a given address

### LINKED ISSUE #566: SIM-BE-Allow getting all transactions for a given address
We need an endpoint to be able to get all transactions for a given address, particularly for the Intelligent Oracles, where we want to be able to see all the transactions that an Oracle has processed

AgustinRamiroDiaz:
Fixes #566

# What

Described in issue

# Why

Described in issue

# Testing done

Manually tested by calling the endpoint with insomnia

# Checks

- [x] I have tested this code

# Reviewing tips

none

# User facing release notes

There's a new RPC endpoint `sim_getTransactionsForAddress` to allow you to get all transactions for a given address

### DIFF
--- a/backend/database_handler/transactions_processor.py
+++ b/backend/database_handler/transactions_processor.py
@@ -1,13 +1,20 @@
 # consensus/services/transactions_db_service.py
+from enum import Enum
 import rlp
 
 from .models import Transactions, TransactionsAudit
 from sqlalchemy.orm import Session
-from sqlalchemy import or_
+from sqlalchemy import or_, and_
 
 from .models import TransactionStatus
 from eth_utils import to_bytes, keccak, is_address
 import json
+
+
+class TransactionAddressFilter(Enum):
+    ALL = "all"
+    TO = "to"
+    FROM = "from"
 
 
 class TransactionsProcessor:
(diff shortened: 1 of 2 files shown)


backend/database_handler/transactions_processor.py
```diff
@@ -182,3 +183,19 @@ def get_transaction_count(self, address: str) -> int:
...
+        transactions = (
+            self.session.query(Transactions)
+            .filter(
+                or_(
+                    Transactions.from_address == address,
+                    Transactions.to_address == address,
+                )
+            )
+            .all()
+        )
```
AgustinRamiroDiaz, line 197: I'm still thinking if we should only check for `to_address`, since that would give us all the ordered `nonce`s and it's probably what we want. WDYT?

cristiam86: in the explorer you get both at the same time sorted by date. Maybe you can include an optional parameter to filter by only_from or only_to

AgustinRamiroDiaz: @cristiam86 I've added the filter :+1:

cristiam86 (approved):

PR #573, opened 2024-10-22
feat: add rollup transaction db table

### LINKED ISSUE #572: Refactor audits table
We need to refactor the current audits database table doing the following:
1. Rename it to "rolllup_transactions"
2. Every time a new GenLayer transaction is created, insert a rollup transaction record with its own data. More info: 
https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_sendtransaction
https://polygonscan.com/tx/0xc24a099f3a4c223cf70f9720796d90bbc00f7ba089c7bc4f4bf2b1bd26a4d58f

Consider the following:
- Nonce can be the timestamp in milliseconds. 
- Gas values can be all zero
- Input data should be the encoded GenLayer transaction

kstroobants:
Fixes #572 

# What

- Add migration file to create rollup_tractions table in database
- Create rollup transaction every time the state of the genlayer transaction changes

# Why

- Needed for appeals

# Testing done

- tested the new feature

# Decisions made

# Checks

- [x] I have tested this code

# Reviewing tips

Check the database


backend/database_handler/migration/versions/1ecaa2085aec_rollup_transactions.py
```diff
@@ -0,0 +1,40 @@
...
+# revision identifiers, used by Alembic.
+revision: str = '1ecaa2085aec'
+down_revision: Union[str, None] = 'ab256b41602a'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('rollup_transactions',
```
cristiam86, line 23: @kstroobants we also need to remove the audits table

backend/database_handler/models.py
```diff
@@ -117,18 +118,33 @@ class Transactions(Base):
...
 
 
-class TransactionsAudit(Base):
-    __tablename__ = "transactions_audit"
-    __table_args__ = (PrimaryKeyConstraint("id", name="transactions_audit_pkey"),)
+class RollupTransactions(Base):
+    __tablename__ = "rollup_transactions"
+    __table_args__ = (PrimaryKeyConstraint("id", name="rollup_transactions_pkey"),)
 
     id: Mapped[int] = mapped_column(Integer, primary_key=True, init=False)
```
cristiam86, line 125: @kstroobants the ID should be the transaction hash

backend/database_handler/transactions_processor.py
```diff
@@ -158,12 +158,16 @@ def insert_transaction(
...
-        transaction_audit_record = TransactionsAudit(
-            transaction_hash=new_transaction.hash,
-            data=self._parse_transaction_data(new_transaction),
+        rollup_transaction_record = RollupTransactions(
+            from_=from_address,
+            to_=to_address,
+            gas=0,
+            gas_price=0,
+            value=value,
+            input=self._transaction_data_to_str(self._parse_transaction_data(new_transaction)),
```
cristiam86, line 167: @kstroobants here we can store the exact same encoded call data that we receive from the frontend

cristiam86 (changes requested):

backend/database_handler/migration/versions/1ecaa2085aec_rollup_transactions.py
kstroobants: fixed in eca17e5fbf511c790d2025babe11cdf8c3bec6b7

backend/database_handler/models.py
kstroobants: Fixed in eca17e5fbf511c790d2025babe11cdf8c3bec6b7. Replaced id by transaction_hash.
It cannot be the transaction_hash from the genlayer because then the table contains multiple entries with the same primary key when adding rollup transactions for each consensus state change and that is not allowed.
For now, I used _generate_transaction_hash() to make a transaction_hash for the rollup_transaction. Maybe it needs to be changed because there is no type.

backend/database_handler/transactions_processor.py
kstroobants: @cristiam86 Do you expect a fix? new_transaction.data contains the call_data. Hence, rollup_transaction also contains it.

cristiam86 (approved):

cristiam86:
@kstroobants good job, please ensure that the tests pass :)

cristiam86 (approved):
Pleas update with main, test, and merge!

cristiam86 (approved):

PR #585, opened 2024-11-06
feat: allow configuring validators from envvar

### LINKED ISSUE #567: SIM-Deploy-Allow configuring validators from a file
For new deployments, it's necessary to create validators in the Simulator
This process ATM is manually done

We should handle the automatic creation of validators from a JSON file (if no validators exist)

AgustinRamiroDiaz:
Fixes #567

# What

Allows configuring validators from the envvar `VALIDATORS_CONFIG_JSON`

# Why

To make deployment easier

# Testing done

- unit tests
- tested with `VALIDATORS_CONFIG_JSON = '[{"stake": 100, "provider": "openai", "model": "gpt-4o", "amount": 2}]'` and `VALIDATORS_CONFIG_JSON = ''`

# Decisions made

- I did it in the same component as the jsonrpc server so that we don't need a new component, which would make deployment more complex
- I didn't use threads to do it since it doesn't take much time, and also threading would've made it more complex given that exception handling across threads is hard

# Checks

- [x] I have tested this code

# Reviewing tips

Read unit tests first to understand the behavior

# Extra

Maybe this feature is useful in the genlayer CLI? @denishacquin @cristiam86 

# User facing release notes

Now you can easily start up your validators from the envvar `VALIDATORS_CONFIG_JSON`

### DIFF
--- a/tests/integration/test_llm_providers_registry.py
+++ b/tests/integration/test_llm_providers_registry.py
@@ -1,16 +1,41 @@
 from tests.common.request import payload, post_request_localhost
 from tests.common.response import has_success_status
 
+# provider, model, plugin
+
 
 def test_llm_providers():
+    providers_and_models_response = post_request_localhost(
+        payload("sim_getProvidersAndModels")
+    ).json()
+    assert has_success_status(providers_and_models_response)
+    providers_and_models = providers_and_models_response["result"]
+
+    gpt4o_provider_id = next(
+        (
+            provider["id"]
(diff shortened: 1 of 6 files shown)


tests/integration/test_llm_providers_registry.py
AgustinRamiroDiaz, line 1: e2e tests were broken prior to this PR

denishacquin (approved):

PR #586, opened 2024-11-07
feat: Added nginx webserver

### LINKED ISSUE #587: Add reverse proxy, make deployments more reliable and secure
### Summary

Add an NGINX server as a reverse proxy for frontend and backend containers, enable automatic container restart, and improve container security.

### Background

Currently, the frontend and backend containers operate without a reverse proxy, which can limit scalability and load balancing. Additionally, there is no automated mechanism for container restart upon failure, which could lead to downtime. Finally, there is an opportunity to improve security by preventing privilege escalation within containers.

### Goals

1. **Add NGINX as a Reverse Proxy**: Introduce NGINX to route traffic between the frontend and backend containers, enabling better load distribution and network management.
2. **Enable Automatic Restart Policy**: Set a restart policy for containers to ensure they restart automatically if they fail or if the machine restarts.
3. **Enhance Security with `no-new-privileges=true`**: Prevent privilege escalation within containers by setting the `no-new-privileges` flag.

### Requirements

- **NGINX Setup**: NGINX should sit in front of the frontend and backend containers, routing traffic appropriately and ensuring high availability.
- **Restart Policy**: Containers should automatically restart on failure or host machine reboot to maintain uptime.
- **Security Update**: Apply the `no-new-privileges=true` security setting to prevent any unauthorized privilege escalations.

### Additional Information

This update aims to improve the robustness, scalability, and security of the containerized application environment.

YeagerAI-Bulat:
Fixes #587
# What 
- Added an NGINX server as a reverse proxy in front of the frontend and backend containers for improved load balancing and resource management.
- Configured a restart policy for containers to automatically restart in case of failure or machine reboot, enhancing application reliability and uptime.  
- Enhanced container security by setting `no-new-privileges=true`, preventing any privilege escalation in the container.
- Implemented JSON-RPC method filtering using nginx+lua to restrict access to sensitive Validator configuration methods when running in production:
  - Added lua scripting support to the NGINX reverse proxy
  - Created a lua script to inspect the JSON-RPC request body and filter requests based on the method name
  - Configured NGINX to execute the lua filter script for POST requests to the backend JSON-RPC endpoint
  - Blocked access to Validator config methods like `sim_deleteAllValidators` and `sim_clearDbTables` in production
# Why
- To introduce NGINX for handling requests to the frontend and backend containers, which helps improve scalability and manage network traffic. It will listen on ports 80 (always redirects to port 443), 443 (frontend with TLS encryption enabled) and 8443 (backend)
- To ensure that containers automatically restart if they fail or if the host machine is restarted, reducing downtime without manual intervention.
- To increase security within the container environment by preventing privilege escalation, aligning with security best practices. 
- To enhance security of the Validator by preventing unauthorized modification of sensitive configuration via the JSON-RPC API when running in production environments. JSON-RPC method filtering allows blocking access to Validator config altering methods.
# Testing Done
- Verified that the NGINX server correctly routes traffic to the frontend and backend. 
- Tested the restart policy by simulating container and machine failures to confirm containers restart as expected.
- Confirmed the `no-new-privileges` security flag is set and functioning as intended to prevent privilege escalation.
- Tested the JSON-RPC method filter:
  - Sent JSON-RPC requests for allowed methods and verified they were proxied to the backend
  - Sent JSON-RPC requests for blocked configuration methods and confirmed the requests were rejected by NGINX with a 400 Forbidden status
# Decisions Made  
- Chose to use NGINX as the reverse proxy server for its compatibility and flexibility in handling web traffic.
- Decided on a `no-new-privileges` security setting to align with security standards in containerized applications.
- Used nginx+lua for JSON-RPC method filtering to leverage NGINX's built-in scripting capabilities and avoid adding another component. Lua provides good performance for inspecting request bodies.
# Checks
- [x] I have tested this code
# Reviewing Tips 
- Reviewers should check the NGINX configuration to ensure correct routing between frontend and backend. Verify JSON-RPC requests are proxied to the backend service.
- Ensure the restart policy works correctly across different environments and conditions.
- Verify that `no-new-privileges=true` is applied correctly in container configuration.
- Check the nginx lua script to confirm it properly filters JSON-RPC requests based on the method. Test with allowed and blocked methods.
# User-Facing Release Notes
- Added NGINX server for improved handling of requests to frontend and backend services. 
- Containers now automatically restart upon failure or machine reboot, minimizing downtime.
- Improved container security by enforcing no privilege escalation with `no-new-privileges=true`. 
- Sensitive JSON-RPC methods for modifying Validator configuration are now blocked in production environments for enhanced security.


nginx.conf
```diff
@@ -0,0 +1,81 @@
+events {
+    worker_connections 1024;
+}
+
+# frontend
+http {
+    server {
+        listen 80;
+        server_name sim.genlayer.com;
```
cristiam86, line 9: @YeagerAI-Bulat we probably need to change this to the new subdomain

Bulat-Gumerov: Retaining the current simulator URL for compatibility as we plan to deploy additional simulators in the future.
Refer to this Jinja template: https://github.com/yeagerai/genlayer-simulator-infra/blob/f7fb254492d52bb692e55dca886b7a9525579337/roles/genlayer/templates/nginx.j2#L9

YeagerAI-Bulat: Retaining the current simulator URL for compatibility as we plan to deploy additional simulators in the future.
Refer to this Jinja template: https://github.com/yeagerai/genlayer-simulator-infra/blob/f7fb254492d52bb692e55dca886b7a9525579337/roles/genlayer/templates/nginx.j2#L9

docker-compose.yml
```diff
@@ -1,4 +1,32 @@
 services:
+  nginx:
+    image: nginx:alpine
```
AgustinRamiroDiaz, line 3: pin this image version please

docker-compose.yml
```diff
@@ -1,4 +1,32 @@
...
+    cap_add:
+      - NET_BIND_SERVICE
+    ports:
+      - "80:80"
+      - "443:443"
+      - "8443:8443"
```
AgustinRamiroDiaz, line 9: we shouldn't clutter the host's ports. Can we use only "weird" ports like 8443 and 2096 maybe?

Also please add a reference to https://developers.cloudflare.com/fundamentals/reference/network-ports/ so we know why we have this limitation

ssl/README.md
AgustinRamiroDiaz, line 1: This description is too scarce for anyone to actually implement this. Please explain further how to fill this directory

docker-compose.yml
```diff
@@ -14,17 +42,25 @@ services:
...
-      target: debug
+      target: ${BACKEND_BUILD_TARGET:-prod}
     environment:
       - FLASK_SERVER_PORT=${RPCPORT}
-      # TODO: remove this in production
-      - PYTHONUNBUFFERED=1
-      - RPCDEBUGPORT=${RPCDEBUGPORT}
+      # # TODO: remove this in production
+      # - PYTHONUNBUFFERED=1
+      # - RPCDEBUGPORT=${RPCDEBUGPORT}
```
AgustinRamiroDiaz, line 63: we should probably use an envvar to set/unset debug mode. As it is now, you've blocked us on development. Please change

nginx.conf
```diff
@@ -0,0 +1,81 @@
...
+
+# backend (jsonrpc)
+    server {
+        listen 8443 ssl;
+        http2 on;
+        server_name sim.genlayer.com;
+
+        ssl_certificate /etc/nginx/ssl/sim.genlayer.com.crt;
+        ssl_certificate_key /etc/nginx/ssl/sim.genlayer.com.key;
+        ssl_client_certificate /etc/nginx/ssl/cloudflare.crt;
```
AgustinRamiroDiaz, line 57: all of these names should also be configurable, or at least explained in the README.md

nginx.conf
```diff
@@ -0,0 +1,81 @@
...
+    server {
+        listen 80;
+        server_name sim.genlayer.com;
+        return 301 https://$server_name$request_uri;
+    }
+
+    server {
+        listen 443 ssl;
+        http2 on;
+        server_name sim.genlayer.com;
```
AgustinRamiroDiaz, line 16: the `server_name` should be configurable (probably through environment variables) on all of these `server`s, since it's something that will very likely change on each deployment.

AgustinRamiroDiaz (changes requested):
Please review all my comments. When implementing the changes, please think also about the development flow, and that this component will be running on all user's machines :smiley: 

Also, could you document how to interact with the nginx component? like which ports and paths to use

docker-compose.yml
YeagerAI-Bulat: These are the default ports for HTTP (port 80) and HTTPS (port 443) web servers. Web browsers use these ports by default.

docker-compose.yml
YeagerAI-Bulat: I've updated the NGINX image to a third-party image that includes Lua support for filtering JSON-RPC methods.

docker-compose.yml
YeagerAI-Bulat: I've added `SERVER_NAME` env var

nginx.conf
YeagerAI-Bulat: I've added `SERVER_NAME` env var

docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 services:
   nginx:
-    image: nginx:alpine
+    image: fabiocicerchia/nginx-lua:1.27.2-almalinux9.4-20240923
```
AgustinRamiroDiaz, line 3: Nit: can we use our own Dockerfile simliar to [this one](https://stackoverflow.com/a/55994069)?

AgustinRamiroDiaz (changes requested):
The first thing I get when trying this out is 

```
Error response from daemon: driver failed programming external connectivity on endpoint genlayer-simulator-nginx-1 (501e24ae5248621e1f2d82f6f6d1ae99aa57f30fc2bb0ccc28e9d095b2375828): failed to bind port 0.0.0.0:80/tcp: Error starting userland proxy: listen tcp4 0.0.0.0:80: bind: address already in use
```

I think a good solution would be to set the nginx service to be run only when a profile like `deploy` is set

.env.example
```diff
@@ -1,71 +1,71 @@
...
-# Front end container details
-VITE_JSON_RPC_SERVER_URL  = 'http://127.0.0.1:4000/api' # if VITE_PROXY_ENABLED = 'true' change to '/api'
-VITE_WS_SERVER_URL   = 'ws://127.0.0.1:4000' # if VITE_PROXY_ENABLED = 'true' change to '/'
-VITE_PLAUSIBLE_DOMAIN = 'simulator.genlayer.com'
-FRONTEND_PORT   = '8080'
-FRONTEND_BUILD_TARGET = 'final' # change to 'dev' to run in dev mode
+# Frontend Configuration
+VITE_JSON_RPC_SERVER_URL = 'http://127.0.0.1:4000/api'  # if VITE_PROXY_ENABLED = 'true' change to '/api'
+VITE_WS_SERVER_URL      = 'ws://127.0.0.1:4000'        # if VITE_PROXY_ENABLED = 'true' change to '/'
+VITE_PLAUSIBLE_DOMAIN   = 'sim.genlayer.com'
```
AgustinRamiroDiaz, line 61: I think it's now 
```suggestion
VITE_PLAUSIBLE_DOMAIN   = 'studio.genlayer.com'
```

.env.example
```diff
@@ -1,71 +1,71 @@
...
-VITE_PROXY_ENABLED = 'false'
+# Vite Proxy Configuration
+VITE_PROXY_ENABLED             = 'false'
 VITE_PROXY_JSON_RPC_SERVER_URL = 'http://jsonrpc:4000'
-VITE_PROXY_WS_SERVER_URL = 'ws://jsonrpc:4000'
+VITE_PROXY_WS_SERVER_URL       = 'ws://jsonrpc:4000'
 
-FRONTEND_BUILD_TARGET = 'final' # change to 'dev' to run in dev mode
+# NGINX Server Configuration
+SERVER_NAME = 'sim.genlayer.com'
```
AgustinRamiroDiaz, line 71:
```suggestion
SERVER_NAME = 'studio.genlayer.com'
```

AgustinRamiroDiaz (changes requested):

.env.example
```diff
@@ -1,73 +1,73 @@
...
-RPCPORT             = '4000'
-# (the port debugpy is listening on)
-RPCDEBUGPORT        = '4678'
+# JsonRPC Server Configuration
+RPCPROTOCOL             = 'http'
+RPCHOST                 = 'jsonrpc'
+RPCPORT                 = '4000'
+RPCDEBUGPORT            = '4678'      # debugpy listening port
+ENVIRONMENT             = 'dev'       # dev/prod
+JSONRPC_REPLICAS        = '1'         # number of JsonRPC container replicas to run, used to scale up for production
```
AgustinRamiroDiaz, line 19: Don't make this configurable, jsonrpc is not prepared to have multiple replicas

docker-compose.yml
```diff
@@ -1,4 +1,35 @@
...
+      timeout: 10s
+      retries: 3
+    restart: always
+    security_opt:
+      - "no-new-privileges=true"
+    logging:
+      driver: "json-file"
+      options:
+        max-size: "10m"
+        max-file: "3"
```
AgustinRamiroDiaz, line 31: this service needs to be off by default, so it doesn't interfere with local installations

docker-compose.yml
```diff
@@ -42,6 +81,22 @@ services:
...
+    deploy:
+      replicas: ${JSONRPC_REPLICAS:-1}
```
AgustinRamiroDiaz, line 99: this needs to be always 1

docker/Dockerfile.nginx
```diff
@@ -0,0 +1,5 @@
+# This Dockerfile is used to build the nginx image with lua and LuaRocks installed
+FROM fabiocicerchia/nginx-lua:1.27.2-almalinux9.4-20240923
```
AgustinRamiroDiaz, line 2: can we use the original nginx image?

.env.example
```diff
@@ -1,73 +1,73 @@
...
-RPCHOST             = 'jsonrpc'
-RPCPORT             = '4000'
-# (the port debugpy is listening on)
-RPCDEBUGPORT        = '4678'
+# JsonRPC Server Configuration
+RPCPROTOCOL             = 'http'
+RPCHOST                 = 'jsonrpc'
+RPCPORT                 = '4000'
+RPCDEBUGPORT            = '4678'      # debugpy listening port
+ENVIRONMENT             = 'dev'       # dev/prod
```
AgustinRamiroDiaz, line 18: where is this being used?

AgustinRamiroDiaz (changes requested):
please rebase with main, and check that there's a new envvar `VITE_IS_HOSTED` that you should take into account

docker-compose.yml
YeagerAI-Bulat: I've turned it off by default, use `docker-compose --profile studio up` to start it explicitly

docker-compose.yml
YeagerAI-Bulat: It is set to 1 replica by default, could be scaled up using this env var

.env.example
YeagerAI-Bulat: I am using docker compose profile for switching the nginx on/off instead of this env var

YeagerAI-Bulat:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=yeagerai_genlayer-simulator&pullRequest=586) **Quality Gate failed**
> 
> Failed conditions  [1 Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=yeagerai_genlayer-simulator&pullRequest=586&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=yeagerai_genlayer-simulator&pullRequest=586)
We need root access to bind on privileged ports - 80, 443
https://security.stackexchange.com/questions/214581/why-is-it-okay-to-run-nginx-or-apache-web-server-as-root-on-port-80-443-and-not

docker/Dockerfile.nginx
AgustinRamiroDiaz: I know Nginx doesn't come with Lua by default. But isn't that the reason there's `RUN luarocks install lua-cjson` below? I don't get it, why don't we use official NGINX image and just add the lua package?

nginx/ssl/README.md
AgustinRamiroDiaz, line 1: The TLS certificates actually need to be in a very particular format as per

```
        ssl_certificate /etc/nginx/ssl/genlayer.com.crt;
        ssl_certificate_key /etc/nginx/ssl/genlayer.com.key;
        ssl_client_certificate /etc/nginx/ssl/cloudflare.crt;
```

.env.example
```diff
@@ -1,64 +1,68 @@
...
 
-# JsonRPC server details
+# JsonRPC Server Configuration
 RPCPROTOCOL         = 'http'
 RPCHOST             = 'jsonrpc'
 RPCPORT             = '4000'
-# (the port debugpy is listening on)
-RPCDEBUGPORT        = '4678'
+RPCDEBUGPORT        = '4678'      # debugpy listening port
+JSONRPC_REPLICAS    = '1'         # number of JsonRPC container replicas to run, used to scale up for production
```
AgustinRamiroDiaz, line 18: This is not being used
```suggestion
```

.env.example
```diff
@@ -1,64 +1,68 @@
...
+# Vite Proxy Configuration (for local development)
+VITE_PROXY_ENABLED             = 'false'
+VITE_PROXY_JSON_RPC_SERVER_URL = 'http://jsonrpc:4000'
+VITE_PROXY_WS_SERVER_URL       = 'ws://jsonrpc:4000'
 
-# Front end container details
-VITE_JSON_RPC_SERVER_URL  = 'http://127.0.0.1:4000/api' # if VITE_PROXY_ENABLED = 'true' change to '/api'
-VITE_WS_SERVER_URL   = 'ws://127.0.0.1:4000' # if VITE_PROXY_ENABLED = 'true' change to '/'
-VITE_PLAUSIBLE_DOMAIN = 'studio.genlayer.com'
-VITE_IS_HOSTED=false
```
AgustinRamiroDiaz, line 55: Do not delete this
```suggestion
VITE_IS_HOSTED=false
```

docker-compose.yml
```diff
@@ -1,4 +1,36 @@
 services:
+  nginx:
+    profiles: ["studio"]  # For hosted studio, turned off by default
```
AgustinRamiroDiaz, line 3:
```suggestion
  nginx:
    environment:
      SERVER_NAME: $SERVER_NAME
    profiles: ["studio"]  # For hosted studio, turned off by default
```

docker-compose.yml
```diff
@@ -45,6 +85,22 @@ services:
         condition: service_healthy
     expose:
       - "${RPCPORT}"
+    healthcheck:
+      test: ["CMD", "python", "backend/healthcheck.py", "--port", "${RPCPORT}"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 30s
```
AgustinRamiroDiaz, line 93: This is not necessary anymore
```suggestion
```

docker-compose.yml
AgustinRamiroDiaz: Remove this, it's not being used and doesn't make sense

AgustinRamiroDiaz (changes requested):

AgustinRamiroDiaz (changes requested):
It works locally now, thanks for fixing that with the profiles

There are still many small changes that need to be addressed. Please read all my comments carefully. 
- This is like the 4th time I'm reviewing this, and many things I have repeated. Please re-read all my comments
- Prior to requesting my review, it's a good practice to review the PR yourself, you'll be surprised how many things you'll find, and it will make the PR review process smoother

.env.example
YeagerAI-Bulat: But we are planning to scale up JSON-RPC containers in the future to handle multiple users in the studio

docker/Dockerfile.nginx
YeagerAI-Bulat: The official Nginx image is lean and lacks prebuilt modules for non-standard features like Lua. Adding Lua to it typically involves either recompiling Nginx with ngx_http_lua_module or using OpenResty (a Lua-focused Nginx distribution). This process increases the complexity and can introduce potential incompatibilities. The fabiocicerchia/nginx-lua image is specifically designed with Lua modules already integrated and tested. It avoids the manual configuration steps needed to compile and add Lua support into Nginx yourself, which can be tedious and error-prone.

AgustinRamiroDiaz (approved):

AgustinRamiroDiaz:
@cristiam86 I haven't tested, but it shouldn't since it's a new container which should only run if the profile is set
The guys from spheron weren't doing orchestration with docker-compose, they were doing it on their own platform, so they only depended on the docker images

PR #589, opened 2024-11-07
feat: 574 appeals implement finality window

### LINKED ISSUE #574: Appeals: implement finality window
Right now the transactions go from accepted to finalized immediately. We need to implement the appeal window.
Its duration should be 1h by default but it could be changed in the env variables.
The transactions in accepted state should stay there for the time set as the appeal window and become final after the time passes.

kstroobants:
Fixes #574 

# What

- Added constant for finality window in base.py (would be nice to make it changeable with CLI like the number of validators)
- Added an accepted queue with its own thread
- Added appeal and timestamp accepted to the transaction model
- Added a button in the frontend to change the transaction in the database
- Added session.commit() whenever a transaction status changes so that the frontend if aware
- Added a testcase to check the appealing process

# Why

- to add more value to the user to explore the consensus mechanism

# Testing done

- tested the new feature

# Decisions made

# Checks

- [x] I have tested this code

# Reviewing tips

It is based on PR #573 (if that is not merged yet, I can edit the base to see the changes more clearly).
Press the button in the simulator (change the time to some seconds instead of 30 minutes)
Run the test

# User facing release notes


backend/consensus/base.py
```diff
@@ -284,43 +317,87 @@ async def exec_transaction(
...
+
+        ConsensusAlgorithm.dispatch_transaction_status_update(
+            transactions_processor,
+            transaction.hash,
+            TransactionStatus.REVEALING,
+            self.msg_handler,
+        )
+        transactions_processor.create_rollup_transaction(transaction.hash)
+
+        time.sleep(2)  # remove this
```
cristiam86, line 368: should we remove it now?

backend/consensus/base.py
```diff
@@ -332,30 +409,31 @@ async def exec_transaction(
...
-        ConsensusAlgorithm.dispatch_transaction_status_update(
-            transactions_processor,
-            transaction.hash,
-            TransactionStatus.FINALIZED,
-            msg_handler,
-        )
+            contract_snapshot.update_contract_state(leader_receipt["contract_state"])
 
         # Finalize transaction
+        consensus_data["final"] = True
```
cristiam86, line 415: It seems that this property is not used and it could be removed.

backend/database_handler/chain_snapshot.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, session: Session):
         self.all_validators = self.validators_registry.get_all_validators()
         self.pending_transactions = self._load_pending_transactions()
         self.num_validators = len(self.all_validators)
+        self.accepted_transaction = self._load_accepted_transactions()
```
cristiam86, line 21:
```suggestion
        self.accepted_transactions = self._load_accepted_transactions()
```

backend/protocol_rpc/endpoints.py
```diff
@@ -623,3 +629,7 @@ def register_all_rpc_endpoints(
         partial(get_transactions_for_address, transactions_processor, accounts_manager),
         method_name="sim_getTransactionsForAddress",
     )
+    register_rpc_endpoint(
+        partial(set_transaction_appeal, transactions_processor),
+        method_name="sim_setTransactionAppeal",
```
cristiam86, line 634:
```suggestion
        method_name="sim_appealTransaction",
```

cristiam86 (changes requested):

backend/consensus/base.py
```diff
@@ -2,12 +2,15 @@
 
 DEFAULT_VALIDATORS_COUNT = 5
 DEFAULT_CONSENSUS_SLEEP_TIME = 5
+DEFAULT_FINALITY_WINDOW = 30 * 60  # 30 minutes
```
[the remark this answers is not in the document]
kstroobants: Solved in eb5ea52a2d4433e9b5e2441d05bf6f027712d486

backend/consensus/base.py
kstroobants: Removed in eb5ea52a2d4433e9b5e2441d05bf6f027712d486

backend/database_handler/chain_snapshot.py
kstroobants: Solved in eb5ea52a2d4433e9b5e2441d05bf6f027712d486

backend/database_handler/transactions_processor.py
```diff
@@ -152,18 +155,15 @@ def insert_transaction(
                 if triggered_by_hash
                 else None
             ),
+            appeal=False,
```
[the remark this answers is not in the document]
kstroobants: Solved in eb5ea52a2d4433e9b5e2441d05bf6f027712d486

backend/protocol_rpc/endpoints.py
kstroobants: Solved in eb5ea52a2d4433e9b5e2441d05bf6f027712d486

backend/consensus/base.py
```diff
@@ -471,6 +518,44 @@ def dispatch_transaction_status_update(
...
+                            self.finalize_transaction(
+                                transaction,
+                                transactions_processor,
+                                lambda contract_address, session=session: ContractSnapshot(
+                                    contract_address, session
+                                ),
+                            )
+                            session.commit()
+                    else:
+                        transactions_processor.set_transaction_appeal(
```
cristiam86, line 549: @kstroobants could you please add some comments here to explain the logic? for example, why do you set the Appeal as false when the transaction was appealed?

backend/consensus/base.py
```diff
@@ -335,18 +339,62 @@ async def exec_transaction(
...
+            transactions_processor,
+            transaction.hash,
+            TransactionStatus.REVEALING,
+            self.msg_handler,
+        )
+        transactions_processor.create_rollup_transaction(transaction.hash)
+
+        time.sleep(2)  # remove this
+
+        transactions_processor.set_transaction_timestamp_accepted(transaction.hash)
```
cristiam86, line 367: @kstroobants can we do this set_transaction_timestamp_accepted inside the dispatch_transaction_status_update?

tests/unit/consensus/test_base.py
```diff
@@ -1,6 +1,9 @@
 from collections import defaultdict
```
cristiam86, line 1: @kstroobants can we abstract the boilerplate code on these tests to another file so it is clear what we are testing?

cristiam86 (changes requested):

backend/consensus/base.py
kstroobants: If it is called dispatch_transaction_update then maybe yes. I could add a flag in the arguments to update the timestamp or not update it.

But actually, I want to keep them separate because I don't update this timestamp a lot compared to the amount of times I call dispatch. It makes it more complicated for the reader of the code if I would combine them.

backend/consensus/base.py
kstroobants: I added a lot of comments to this file in this commit https://github.com/yeagerai/genlayer-studio/pull/637/commits/27624bbf7bb6a3935352e45c89160d6c0428e06a.

Yes, it did not make sense to put it on False. It is solved in next PRs.

tests/unit/consensus/test_base.py
kstroobants: Done in #43da57eda8aad30fb3c2144d5930e1952e986d66

cristiam86 (approved):

PR #592, opened 2024-11-11
feat: rename simulator to studio and handle hosted environment

### LINKED ISSUE #590: SIM-Studio renaming and exceptions
We need to:

- Hide UI related to adding validators when on hosted environment
- Adapt the routing to remove the simulator subpath `https://studio.genlayer.com/simulator/run-debug`
- Gatekeep or disable sensitive endpoints
- Update the plausible domains to studio.genlayer.com (both in codebase + in plausible)
- Adapt copy where necessary
- Update the Docs

denishacquin:
Fixes #590 
# What

- Hid all UI related to adding/editing/deleting validators when on hosted environment
- Adapt the routing to remove the simulator subpath https://studio.genlayer.com/**simulator**/contracts -> https://studio.genlayer.com/contracts
- Updated the plausible domains to studio.genlayer.com (need to update in plausible.com when merge!!)
- Changed `simulator` to `studio` everywhere in the copy
- Normalized the copy to use a capital S for every Studio mention

image
image

# Why

- to account for the renaming and hosting of the studio

# Testing done

- tested the conditional UI for local vs hosted

# Decisions made

- decided to remove the /simulator subpath in the vue router -> made no sense to keep it and it's cleaner this way
- decided to simply detect localhost in order to conditionally display some elements
- added a catch all route to redirect to home path

# Checks

- [x] I have tested this code

# User facing release notes

- Adapted the copy of the simulator to reflect its rebranding to "Studio"


frontend/src/hooks/useConfig.ts
```diff
@@ -0,0 +1,11 @@
+import { ref, computed } from 'vue';
+
+export const useConfig = () => {
+  const isLocalNetwork = ref(window.location.hostname === 'localhost');
```
cristiam86, line 4: @denishacquin I would decide this based on an env variable

cristiam86 (changes requested):
Add the HOSTED flag based on an env variable

frontend/src/hooks/useConfig.ts
denishacquin: That was actually my first intention, I just decided to go a for a quicker solution. But you're absolutely correct. And thanks for the copying tip, wasn't aware of that.

cristiam86 (approved):

PR #595, opened 2024-11-12
feat: adding frontend profile

### LINKED ISSUE #594: Run simulator on headless mode
The goal is that the user can choose to run the Localnet without the frontend UI

Related to: https://github.com/orgs/yeagerai/projects/11/views/3?pane=issue&itemId=78282962&issue=yeagerai%7Cgenlayer-cli%7C72

epsjunior:
Fixes #594

## What

- Added a `frontend` profile to the `frontend` service in the `docker-compose.yml` file.
- Now, users can choose to start the simulator without the `frontend` UI by specifying or omitting the `frontend` profile.

## Why

- This change enables users to run the simulator in **headless mode**, making it more flexible by allowing services to start without the `frontend` UI when not needed. This addresses the need for lightweight, backend-only setups for local testing or headless environments.

## Testing done

- Tested running `docker compose up` to start all services without the `frontend` service.
- Tested `docker compose --profile frontend up` to start all services, including the `frontend` UI.
- Verified compatibility and successful startup on both **Debian** and **macOS** systems.

## Decisions made

- Used the `frontend` profile to simplify managing optional services without needing multiple Compose files or complex configurations.

## Checks

- [x] I have tested this code

## Reviewing tips

- The reviewer can check for proper use of the `frontend` profile in the `frontend` service to verify the service is now optional and can be included or excluded by specifying profiles.

## User facing release notes

- Users can now run the simulator in headless mode, omitting the `frontend` UI by default and enabling it when needed with `docker compose --profile frontend up`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,7 @@
 services:
   frontend:
+    profiles:
+      - frontend
     build:
       context: ./
       dockerfile: ./docker/Dockerfile.frontend


cristiam86:
@epsjunior This would break the current version CLI by not launching the UI. Can we just add a "headless" parameter in the CLI command so, in that case only, the CLI doesn't run the frontend?

PR #618, opened 2024-11-19
fix: fix concurrency issues in tx status update

### LINKED ISSUE #605: SIM-FE hanging of the status in left panel
Simulator frontend: Hanging of the status in left panel and not seeing get function buttons. Sometimes refresh solves it. It occurs more than 50% of the time. I only tried deploying a contract. No idea if it also happens when doing other transactions.

Could be my chrome browser but it should not matter. I cleared also my cache and I still see the issue. Maybe clear your browser's cache if you cannot replicate it.

The issue was added in one of the following commits. I worked on earlier commits where I did not see this:
4f1f948 (tag: v0.20.0, origin/main, origin/HEAD, main) fix: hide methods during deployment to avoid pointing to the wrong contract instance (#582)
771db2a feat: update transactions based on subscriptions events and remove polling (#583)
cb516e4 updated GH actions workflow (#597)
b3e83d0 (tag: v0.19.0) feat: allow configuring validators from envvar (#585)
1e66f9d feat: integrate OpenHands for automatic code generation from GitHub issues (#576) (#578)

wizard of coin stuck:
wizard_of_coin_stuck

wizard of coin resolved after refreshing the browser (0x0319.. is the same and logs show refresh):
wizard_of_coin_reshed_solved

llm erc20 stuck on accepted, same as wizard of coin so it is not contract dependent:
llm_erc20_1_stuck

llm erc20 new deployment, stuck again but on finalized. No get methods visible:
llm_erc20_2_stuck_on_finalized_no_get_methods

llm erc20 refresh of browser shows no deployment, but in list there is a deployed contract (number is same as previous picture):
llm_erc20_2_refresh_of_finalized_shows_no_deployment_but_in_list_there_is_a_deploy

denishacquin:
Fixes #605 

# What

- Added a time tracking system for the fetching of updated tx data to account for rapid websocket events resulting in concurrent requests not always resolving in the right order

# Why

- to fix a bug

# Testing done

- tested the bug fix
- frontend unit OK

# Decisions made

- Initially tried with a deboucing system but didn't manage to make it work, so switched to a timestamp tracking system
- Allowed get transaction endpoint to return null if a tx doesn't exist

# Checks

- [x] I have tested this code

# Reviewing tips

- deploy contracts and fire write transactions; all the statuses should eventually properly end up in **finalized** state

# User facing release notes

- Fixed a bug where the statuses wouldn't update properly in the transaction list

### DIFF
--- a/frontend/src/hooks/useTransactionListener.ts
+++ b/frontend/src/hooks/useTransactionListener.ts
@@ -33,8 +33,6 @@ export function useTransactionListener() {
 
     const newTx = await getTransaction(hash, requestTime);
 
-    if (!newTx) return;
-
     if (!newTx) {
       console.warn('Server tx not found for local tx:', newTx);
       transactionsStore.removeTransaction(newTx);
--- a/frontend/src/hooks/useTransactionListener.ts
+++ b/frontend/src/hooks/useTransactionListener.ts
@@ -1,24 +1,12 @@
 import { useTransactionsStore } from '@/stores';
 import type { TransactionItem } from '@/types';
 import { useWebSocketClient } from '@/hooks';
-import debounce from 'lodash/debounce';
 
 export function useTransactionListener() {
   const transactionsStore = useTransactionsStore();
   const webSocketClient = useWebSocketClient();
   // Track the latest request timestamp for each transaction
   const latestRequestTimes = new Map<string, number>();
-
-  const debouncedGetTransaction = debounce(
-    async (hash: string) => {
-      console.log(`🔄 Fetching transaction ${hash}`);
-      const result = await transactionsStore.getTransaction(hash);
-      console.log(`✅ Fetched status: ${result.status}`);
-      return result;
-    },
-    500,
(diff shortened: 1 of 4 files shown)


frontend/src/hooks/useTransactionListener.ts
```diff
@@ -13,8 +14,24 @@ export function useTransactionListener() {
     );
   }
 
+  async function getTransaction(hash: string, requestTime: number) {
+    const result = await transactionsStore.getTransaction(hash);
+
+    // Only process the result if this is still the latest request for this hash
+    if (latestRequestTimes.get(hash) === requestTime) {
```
cristiam86, line 21: @denishacquin I'm not sure that I understand this. Could you please clarify? it seems a bit odd

denishacquin: @cristiam86 Let's park this PR for now since it should be solved in the backend by an open PR of @kstroobants .

As soon as we merged said PR, let's see if the symptoms still occur and if this fix makes sense still. Moving back to draft.

PR #645, opened 2024-11-25
fix: update genvm: improve error handling, add non comparative equivalence principle, add timestamp

kp2pml30:
Fixes no issue

# What

- updated genvm

# Why

- to fix a bug (`chmod +x`)
- upgradde genvm
- add non comparative eq_principle
- improve logging

# Testing done

- tested the new feature
- tested the bug fix

# Decisions made

# Checks

- [x] I have tested this code

# Reviewing tips

# User facing release notes


backend/node/base.py
```diff
@@ -180,9 +180,54 @@ async def get_contract_data(
...
+                data={
+                    "result": f"{res.result!r}",
+                    "stdout": res.stdout,
+                    "stderr": res.stderr,
+                    "genvm_log": res.genvm_log,
+                },
+            )
+        )
+
+    async def get_contract_schema(self, code: str) -> str | dict:
```
kp2pml30, line 202: we should add a better type here probably, but that is more related to dumping all errors in a more user friendly way in general I think... Because right now on exception only limited information is shown and it's arguments are not split into separate json fields but are rather encoded as a string, which leads to unreadable backslash madness in the ui

examples/contracts/llm_erc20.py
```diff
@@ -34,24 +35,20 @@ def transfer(self, amount: int, to_address: str) -> None:
...
+"""
 
-        final_result = gl.eq_principle_prompt_noncomparative(
-            compute_updated_balances,
-            """
-The new_balance of the sender should have decreased
-in the amount sent and the new_balance of the receiver should have
-increased by the amount sent. Also, the total sum of all balances
-should remain the same before and after the transaction""",
+        final_result = (
```
cristiam86, line 45: @kp2pml30 could you please explain the non-comparative eq principle a bit more? it would also be great to add them to the docs :)

kp2pml30: I am trying to get some feedback from Edgars, as soon as he answers I will add

backend/node/base.py
```diff
@@ -240,6 +274,13 @@ async def _run_genvm(
         snapshot_view = _SnapshotView(
             self.contract_snapshot, self.contract_snapshot_factory, readonly
         )
+
+        # FIXME(core-team): please provide valid date
+        transaction_datetime = datetime.datetime.fromisoformat(
```
kp2pml30, line 279: here if you pass correct `datetime.datetime` everything will work. I am not a high skilled full stack developer, so I left database patching to you, sorry @cristiam86

backend/protocol_rpc/endpoints.py
```diff
@@ -346,7 +347,10 @@ async def get_contract_schema_for_code(
         msg_handler=msg_handler.with_client_session(get_client_session_id()),
         contract_snapshot_factory=None,
     )
-    return json.loads(await node.get_contract_schema(contract_code))
+    schema = await node.get_contract_schema(contract_code)
+    if isinstance(schema, dict):
```
cristiam86, line 351: @kp2pml30 shouldn't this validation be:
```suggestion
    if not isinstance(schema, dict):
```

examples/contracts/llm_erc20.py
cristiam86: @kp2pml30, is this ready, or did you get the feedback that you needed?

backend/node/base.py
cristiam86: @kp2pml30 sorry but I'm not sure what do you mean with database patching. Here we should pass the current timestamp on ISO format in UTC timezone.
Could it be: `transaction_datetime  = datetime.now(timezone.utc)` ?

backend/node/base.py
```diff
@@ -178,8 +178,13 @@ async def get_contract_data(
         from_address: str,
         calldata: bytes,
     ) -> Receipt:
+        now = datetime.now().strftime("%Y-%m-%d %H:%M:%S.%f")
```
kp2pml30, line 181: why not `datetime.now().isoformat()` ?

kp2pml30: It is also incorrect as it passes current time without timezone info. it should be `datetime.now().astimezone(datetime.timezone.utc).isoformat()`

backend/node/base.py
```diff
@@ -266,10 +276,7 @@ async def _run_genvm(
             self.contract_snapshot, self.contract_snapshot_factory, readonly
         )
 
-        # FIXME(core-team): please provide valid date
-        transaction_datetime = datetime.datetime.fromisoformat(
-            "2024-11-28T06:42:42.424242Z"
-        )
+        transaction_datetime = datetime.datetime.fromisoformat(transaction_created_at)
```
kp2pml30:
```py
>>> datetime.datetime.fromisoformat(None)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: fromisoformat: argument must be str
```

cristiam86: This one is weird, probably not appropriately updated because that's not the latest version.
For the None case, we can set a default, but that should be managed when the transaction is created.

backend/node/genvm/base.py
```diff
@@ -136,16 +136,17 @@ async def run_contract(
...
     ) -> ExecutionResult:
         assert date.tzinfo is not None
         message = {
             "is_init": is_init,
             "contract_account": contract_address.as_b64,
             "sender_account": from_address.as_b64,
             "value": None,
-            "datetime": date.isoformat(),
         }
+        if date is not None:
```
kp2pml30, line 148: genvm has a magic default date, so here I just won't place any date if it wasn't given

PR #646, opened 2024-11-26
fix: cross-client event logs on hosted env, tx async data inconsistencies

denishacquin:
# What

- Fixes **event logs sent across users on hosted environment** (the issue appeared with the genvm integration; we were not tracking the client session id on genvm/consensus-related events anymore, effectively dispatching them to all clients. To fix this, I added the transaction hash-scoping to those events as well).
- **Adds a loading state on call method buttons** (wasn't needed before because it was instant; now that the call takes a bit of time it's better for ux responsiveness to set the button to a loading state while the call is processing).
- **Hide provider settings** when on hosted environment
- Fixes **transaction async data errors and inconsistencies** (specifically: handling the case where a server tx doesn't exists anymore after a db reset properly and clearing local txs if their server equivalent doesn't exists anymore; 'reset storage' was only clearing the (example) contracts and not their related client txs - I made sure that those are now removed from the storage as well).

# Why

- To fix some bugs on the hosted environment
- To fix some bugs related to tx async data
- To improve UX

# Testing done

- Tested the fixes and features
- Updated unit tests
- frontend units OK
- frontend e2e OK

# Decisions made

- Added a transaction_hash parameter to the LogEvent class in the backend in order to use it as a room filter for websocket events.

# Checks

- [x] I have tested this code

# Reviewing tips:
- @kp2pml30 can you confirm it's OK to pass transaction_hash as a parameter in the genvm-related methods?

### DIFF
--- a/backend/consensus/base.py
+++ b/backend/consensus/base.py
@@ -323,6 +323,7 @@ class ConsensusAlgorithm:
                     EventType.ERROR,
                     EventScope.CONSENSUS,
                     "Failed to reach consensus",
+                    transaction_hash=transaction.hash,
                 )
             )
             ConsensusAlgorithm.dispatch_transaction_status_update(
@@ -353,6 +354,7 @@ class ConsensusAlgorithm:
                 EventScope.CONSENSUS,
                 "Reached consensus",
                 consensus_data.to_dict(),
+                transaction_hash=transaction.hash,
             )
         )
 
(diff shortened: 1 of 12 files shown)


kp2pml30:
> can you confirm it's OK to pass transaction_hash as a parameter in the genvm-related methods?

yes, sure

kp2pml30 (approved):

cristiam86 (approved):

PR #653, opened 2024-11-27
fix: rework eth_call according to spec

### LINKED ISSUE #621: Rework `eth_call` endpoint
1. it should return contracts return value without transforming it (right now we have `to_string . decode`)
2. it should be able to return jsonrpc error on error, right now it returns Receipt.to_dict with successful status

kp2pml30:
Fixes #621

# What

- now eth_call returns base64 encoded calldata and rpc error on error

# Why

- to fix a bug

# Testing done

- tested the new feature manually
- integrations tests pass (except for one malformed...)

# Decisions made

# Checks

- [x] I have tested this code

# Reviewing tips

# User facing release notes

### DIFF
--- a/frontend/src/components/Simulator/ContractMethodItem.vue
+++ b/frontend/src/components/Simulator/ContractMethodItem.vue
@@ -7,6 +7,7 @@ import { ChevronDownIcon } from '@heroicons/vue/16/solid';
 import { useEventTracking, useContractQueries } from '@/hooks';
 import { unfoldArgsData, type ArgData } from './ContractParams';
 import ContractParams from './ContractParams.vue';
+import * as calldata from '@/calldata';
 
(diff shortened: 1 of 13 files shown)


frontend/src/components/Simulator/ContractMethodItem.vue
```diff
@@ -37,8 +38,13 @@ const handleCallReadMethod = async () => {
       }),
     );
 
-    responseMessage.value =
-      typeof result === 'string' ? result : JSON.stringify(result);
+    let repr: string;
+    if (typeof result === 'string') {
```
kp2pml30, line 42: I think type annotations for callReadMethod are incorrect because it specifies that it returns `{ [key: string]: any }`

cristiam86 (approved):

PR #683, opened 2024-12-05
fix: Revert "feat: detect git tag version and display it in frontend UI and console"

denishacquin:
Reverts yeagerai/genlayer-studio#672


cristiam86:
Fixed in https://github.com/yeagerai/genlayer-studio/pull/645

PR #687, opened 2024-12-05
feat: pushing version when workflow is called using schedule

### LINKED ISSUE #686: Not pushing version when the release workflow is called using schedule

epsjunior:
Fixes #686 

# What

- Implemented logic to dynamically fetch the latest Git tag for use in Docker image tagging.
- Updated the workflow to use the fetched tag 

# Why

- To ensure that Docker images are consistently tagged with the most recent Git tag, improving traceability and alignment with the repository's versioning strategy.
- Adds more flexibility and reliability to the workflow by using Git's tag history.

# Testing done

- Tested the solution by triggering the workflow directly from my `686-not-pushing-version-when-the-release-workflow-is-called-using-schedule` branch using GitHub CLI.
- The successful runs can be observed in the Actions history under this branch.

# Decisions made

- Decided to fetch the most recent Git tag dynamically instead of relying on `package.json` or fixed tags for better traceability.

# Checks

- [x] I have tested this code

# Reviewing tips

- Focus on the `git fetch --tags` and `git describe` steps to ensure they correctly handle fetching and determining the latest tag.
- Verify that the workflow tags the Docker image with both the latest Git tag and `latest`.

# User facing release notes

- Docker images are now automatically tagged with the most recent Git tag and `latest`, ensuring better version traceability and alignment with repository releases.

### DIFF
--- a/.github/workflows/docker-build-and-push-image.yml
+++ b/.github/workflows/docker-build-and-push-image.yml
@@ -31,25 +31,11 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      # TODO: Once we have the 1Password service account token, we can use the following action to load secrets
-      # "better done than perfect"
-
-      # - name: Load secrets from 1Password
-      #   uses: 1password/load-secrets-action/configure@v1
-      #   id: op-load-secret
-      #   with:
-      #     service-account-token: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
-      #   env:
-      #     DOCKERHUB_USERNAME:
-      #     DOCKERHUB_TOKEN:
-
       - name: Login to Docker Hub
         uses: docker/login-action@v3
         with:
           username: ${{ inputs.dockerhub_username }}
           password: ${{ secrets.dockerhub_token }}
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-build-and-push-image.yml
```diff
@@ -31,43 +31,36 @@ jobs:
...
         uses: docker/metadata-action@v5
         with:
           images: |
             ${{ inputs.dockerhub_repo }}
           tags: |
-            type=ref,event=tag
-            type=raw,value=latest
-            type=semver,pattern={{version}}
+            ${{ env.latest_tag }}
+            latest
```
AgustinRamiroDiaz, line 63: nit: what happens if you only added 

```
      - name: Fetch All Tags
        run: git fetch --tags
```
? Maybe the `docker/metadata-action@v5` is able to get the tags in that case.

Anyway, just a nit thinking about how to simplify this

AgustinRamiroDiaz (approved):
Awesome! :rocket:

.github/workflows/docker-build-and-push-image.yml
epsjunior: I searched and apparently it will not work because we are using workflow_call to start this workflow

PR #711, opened 2024-12-10
feat: add number input for finality window

### LINKED ISSUE #710: SIM-FE-Finality window
Add an option to the frontend to change the finality window of the appeal process.

kstroobants:
Fixes #710

# What

- Added a field to the frontend to change the finality window of the appeal process

# Why

# Testing done

- Tested the new feature on the webpage

# Decisions made

I have put the field below the leader. Let me know if it is better to put it in the settings tab.
Screenshot 2024-12-10 at 08 28 30
Screenshot 2024-12-10 at 08 29 00
Screenshot 2024-12-10 at 08 28 48
Screenshot 2024-12-10 at 08 29 14

# Checks

- [x] I have tested this code

# Reviewing tips

See code changes and run the studio.

# User facing release notes

A number field is added to the frontend to change the finality window of the appeal process.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -75,4 +75,4 @@ HEURISTAIAPIKEY         = '<add_your_heuristai_api_key_here>'
 VALIDATORS_CONFIG_JSON = ''
 
 # Consensus mechanism
-FINALITY_WINDOW = 1800 # in seconds
+VITE_FINALITY_WINDOW = 1800 # in seconds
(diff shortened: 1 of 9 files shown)


kstroobants:
@denishacquin Feel free to change the UI look.
@cristiam86 I don't know if this feature is good/bad for the online hosted studio. We should let Bulat know if he should disable access the the rpc method but then the field should also be disabled.

kstroobants:
> @cristiam86 I don't know if this feature is good/bad for the online hosted studio. We should let Bulat know if he should disable access the the rpc method but then the field should also be disabled.

Added check in frontend field based on VITE_IS_HOSTED so it won't be shown in the hosted version. @YeagerAI-Bulat When hosting we should write a meaningful finality window in the .env.example file.

PR #712, opened 2024-12-10
fix: consensus deploy contract when finalized

### LINKED ISSUE #708: Consensus: deploy contract when finalized
Write and get methods should be accessible only when transaction is finalized

kstroobants:
Fixes #708

# What

- Redo the change so that the contract is deployed when the transaction is finalized. I could not find any other issue than the issue where the appeal window loop does not recover, that blocks the transaction to go to the finalized state.

# Why

- To fix a bug

# Testing done

- Tested the bug fix in the localhost and ran the consensus unit tests locally

# Decisions made

# Checks

- [ ] I have tested this code

# Reviewing tips

Run studio and see code changes

# User facing release notes

Contract is deployed when the transaction is in the finalized state after waiting for the finality window. The write methods will be accessible when the transaction is in the finalized state.

### DIFF
--- a/backend/consensus/base.py
+++ b/backend/consensus/base.py
@@ -411,7 +411,6 @@ class ConsensusAlgorithm:
         self,
         transaction: Transaction,
         transactions_processor: TransactionsProcessor,
-        contract_snapshot_factory: Callable[[str], ContractSnapshot],
     ):
         # temporary, reuse existing code
         # and add other possible states the transaction can go to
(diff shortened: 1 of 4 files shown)


kstroobants:
I won't merge this PR before there is a user friendly solution in the main branch to change the finality window in .env or we put the default on a low number.

kstroobants:
Closed because it was implemented correctly.

When a transaction tx1 is sent to deploy an intelligent contract then from the accepted state you are able to submit the next transaction tx2 towards that intelligent contract.

PR #753, opened 2024-12-16
chore: bump genvm version to v0.0.11, add contract deployment from IC

kp2pml30:
Fixes no issue

# What

- updated genvm version
- made code bytes instead of string
- added contract factory support

# Why

- to fix a bug

# Testing done

- [x] by-hand ui testing
- [x] new deploy test
- [x] all old integration tests

# Decisions made

1. Code was changed to bytes to support zips
2. **IMPORTANT** I at least for now changed backend image due to incompatibilities in libc (github now uses ubuntu 24 with glibc 2.39 which is incompatible with debian bookworm), I can try to fix it later with rpath magic in alternative libc installation

# Checks

- [x] I have tested this code

# Reviewing tips

# User facing release notes


webrequest/server.py
```diff
@@ -87,6 +87,7 @@ async def status(x: str) -> dict:
 async def llm_genvm_module_call(
     encoded_model: str,
     prompt: str,
+    mode: str,
```
kp2pml30, line 90: This will be implemented later, it is for json output

backend/consensus/base.py
```diff
@@ -779,14 +779,49 @@ async def handle(self, context):
             nonce = context.transactions_processor.get_transaction_count(
                 context.transaction.to_address
             )
+            data: dict
```
cristiam86, line 782: @kp2pml30 this should happen on transactions accepted

cristiam86 (approved):

cristiam86:
@kp2pml30 do you have an example on how to deploy Intelligent Contracts from another contract?

PR #759, opened 2024-12-17
fix: consensus 2nd tx should not finish before 1st tx

kstroobants:
Fixes #661

# What

- Add check to finalize the transaction only if the previous transaction is finalized

# Why

- To fix a bug

# Testing done

- Tested the bug fix in the studio

# Decisions made

# Checks

- [x] I have tested this code
- [x] I have reviewed my own PR
- [x] I have created an issue for this PR
- [x] I have set a descriptive PR title compliant with [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

# Reviewing tips

Try in the studio

# Depends on
This PR contains code from PR #812

# User facing release notes

The transaction can only be finalized if the previous transaction is finalized.

PR #657, opened 2024-11-28
feat: appeals undetermined transactions

kstroobants:
Fixes #638 

# What

- Added appeal window to undetermined state
- Changed timestamp_accepted and added appeal_undetermined in database
- Activated frontend button in undetermined state and added button also to modal
- Leader only has no appeal button and no appeal window
- Implemented the state transitions to process the appeal
- Added test for leader appeals

# Why

- To add more value to the user

# Testing done

- tests/unit/consensus/test_base.py::test_exec_undetermined_appeal

# Checks

- [x] I have tested this code
- [x] I have reviewed my own PR
- [x] I have created an issue for this PR
- [x] I have set a descriptive PR title compliant with [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

# Reviewing tips

- Run simulator
- Run test_exec_undetermined_appeal
- Check code changes

PR #768, opened 2024-12-19
feat: consensus appeal success should rollback all newer transactions to pending

### LINKED ISSUE #662: Consensus: appeal success should rollback all newer transactions to pending
You can use created_at field from Transactions.

kstroobants:
Fixes #662 

# What

- Added start and stop event for async tasks for each contract based on address.
- Added boolean flag to see if the task is still running for each contract based on address.
- For each contract when a rollback is needed, I empty the pending queue and I put the status of the newer transactions to pending.
- Same applies for accepted and undetermined transactions.
- Add order by created_at in get_pending_transactions() query to process the oldest first.
- The validator set is kept when rollbacked. This means that if a transaction has an increased validator set because of an appeal and that transaction need to be rollbacked to pending then it keeps the increased validator set (confirmed with Kiril).

# Why

- Future transactions depend on each other. If one needs to go back to pending because of a successful appeal then the transactions that came after need to be rolled back.
- Creates value for the user.

# Testing done

- Tested the new feature in the localhost

# Decisions made

- I use asyncio.Event() to stop and start the tasks. Other option was to use a boolean flag.
- With while is_pending_queue_task_running() in rollback_transactions() I let the processing of the current transaction finish (until accepted or undetermined state) by the pending queue processor task to return it back to pending in a clean way. If it needs to be faster we can add a check for the stop event in the state transition while loop of exec_transaction() to stop after the current state transition so it doesn't need to go until accepted/undetermined but we need to see if it doesn't break anything.
- I empty the pending queue as I assume all transactions in the pending queue were send after the appealed accepted/undetermined transaction. So the _crawl_snapshot will get the rolled back transactions and the pending transactions that were already in the queue. Other option is to store the created_at of the appealed transaction in the consensusalgorithm class and only process if the current selected pending transaction's created_at is the same as the stored value. This will filter the newer that we present in the queue and they will be added at the end of the queue by the _crawl_snapshot.
- The leader appeal is processed by the appeal_window thread, implemented in PR #657. Otherwise the code will be confusing when a leader appeal should be rollbacked when a validator appeal is successful (to select the correct validator set).
- I rollback all newer transactions for successful appealed accepted and undetermined transaction to the pending state. Other option is to not rollback when the contract state was not changed.

# Checks

- [x] I have tested this code

# Reviewing tips

Check code changes and try it.

# Depends on
This PR contains code from PR #759 

# User facing release notes

A leader appeal success and validator appeal success will rollback all newer transactions to the pending status.


backend/database_handler/transactions_processor.py
```diff
@@ -382,3 +382,24 @@ def get_transactions_for_block(
...
+    def get_newer_transactions(self, transaction_hash: str):
+        transaction = (
+            self.session.query(Transactions).filter_by(hash=transaction_hash).one()
+        )
+        address = transaction.to_address or transaction.from_address
+        transactions = (
+            self.session.query(Transactions)
+            .filter(
+                Transactions.created_at > transaction.created_at,
+                or_(
```
cristiam86, line 395: @kstroobants only Transaction.to_address

kstroobants: Solved with 87b3b4317fd0cb6da0647ce940f959f23422741f

backend/consensus/base.py
```diff
@@ -152,6 +152,79 @@ def accounts_manager_factory(session: Session):
...
+            transactions_processor (TransactionsProcessor): Instance responsible for handling transaction operations within the database.
+            snapshot (ChainSnapshot): Snapshot of the chain state.
+            accounts_manager (AccountsManager): Manager for accounts.
+            contract_snapshot_factory (Callable[[str], ContractSnapshot]): Factory function to create contract snapshots.
+            node_factory (Callable[[dict, ExecutionMode, ContractSnapshot, Receipt | None, MessageHandler, Callable[[str], ContractSnapshot]], Node]): Factory function to create nodes.
+            msg_handler (MessageHandler): Handler for messaging.
+        """
+        self.transaction = transaction
+        self.transactions_processor = transactions_processor
+        self.snapshot = snapshot
```
cristiam86, line 211: @kstroobants  is this the chain_snapshot? If so, can you please rename it?

cristiam86 (approved):

backend/consensus/base.py
kstroobants: Solved in e4de943b86b7fd1e14f832883d0f87a817df21fd

PR #787, opened 2025-01-03
feat: consensus store acceptance and finalized states on contracts

### LINKED ISSUE #783: Consensus: store acceptance and finalized states on contracts

kstroobants:
Fixes #783 

# What

- Contract contains accepted and finalized state in a states dictionary.
- The encoded state data member of the ContractSnapshot points to the accepted state. This is used in _SnapshotView reads and writes.
- In the transaction accepted state, the contract accepted state is written to the database. The state is taken from the leader_receipt.
- In the transaction finalized state, the contract finalized state is written to the database. The state is taken from the leader_receipt.
- In case a transaction is executed on a contract that does not contain accepted and finalized state, then the old version of the state of an existing contract is updated.
- Updated existing tests.

# Why

- To add more value to the user.

# Testing done

- Tested the new feature in the studio

# Decisions made

The frontend get method show only the accepted contract state. I will create a new issue to show both accepted and finalized contact states.

# Checks

- [x] I have tested this code

# Reviewing tips

- Run the studio.
- Check the code changes.

# User facing release notes

The finalized contact state is tracked.


cristiam86 (changes requested):
@kstroobants this approach doesn't always work: 
- Let's say we have contract C and transactions T1, T2, and T3. 
- Let's say T1 is finalized 
--- At this point, C accepted and finalized state are the ones after T1 (let's name them C-Acc-T1 and C-Fin-T1)
- Let's say T2 and T3 are accepted
--- At this point, C accepted state is after T3 (C-Acc-T3) and finalized state is still after T1 (C-Fin-T1)
- Let's say T3 is appealed successfully and sends T3 back to pending.
--- At this point, C accepted state should be after T2 (C-Acc-T2) and finalized state is still after T1 (C-Fin-T1)

Having said that, we also need the state after every transaction so we can roll back a given contract accepted state to any point in time.

kstroobants:
> @kstroobants this approach doesn't always work:
> 
> * Let's say we have contract C and transactions T1, T2, and T3.
> * Let's say T1 is finalized
>   --- At this point, C accepted and finalized state are the ones after T1 (let's name them C-Acc-T1 and C-Fin-T1)
> * Let's say T2 and T3 are accepted
>   --- At this point, C accepted state is after T3 (C-Acc-T3) and finalized state is still after T1 (C-Fin-T1)
> * Let's say T3 is appealed successfully and sends T3 back to pending.
>   --- At this point, C accepted state should be after T2 (C-Acc-T2) and finalized state is still after T1 (C-Fin-T1)
> 
> Having said that, we also need the state after every transaction so we can roll back a given contract accepted state to any point in time.

Solved in PR #799

cristiam86 (approved):

Issue #796, opened 2025-01-03
Consensus: appeals use wrong contract state


kstroobants:
Same for validator and leader appeal.

The extra validators in the appeal are using the latest contract state but they should use the same contract state that the previous validators and leader were using.

_SnapshotView class should contain the same contract snapshot. It is used when calling node.exec_transaction().

Example:
Imagine the contact code contains
```
# write method
@gl.public.write
def update_storage(self, new_storage: str) -> None:
    self.storage += new_storage
```

Tx1: Contract is deploy and self.storage = "a" -> database storage contains "a"
Tx2: call write method update_storage with "b" -> database storage contains "ab"

The storage in the contract state used in appealing Tx2 should be "a" and not "ab" before executing the update_storage. Otherwise the validators will have "abb".

PR #799, opened 2025-01-06
feat: consensus appeals use wrong contract state and restore contract state after appeal successful

### LINKED ISSUE #790: Consensus: restore contract state after appeal successful
Look for the most recent accepted transaction state after execution and restore that value.

Only for validator appeal because the transaction is calculated again starting from Pending.
For the leader appeal, the transaction goes directly to the Accepted state where the contract state is registered/updated (no recalculation).

kstroobants:
Fixes #790 and #796

# What

- Add contact_snapshot field to the Transactions database  table.
- contact_snapshot field is set in accepted state.
- The transaction.contract_snapshot is set to None in Pending state for rollbacks (not for leader appeal).
- When a validator appeal failed, the contract state is not again updated/deployed.
- contact_snapshot field contains the contract state before evaluating the transaction. This contract state is used to restore the database's contract state when a validator appeal is successful and the transactions are rolled back. There is no need to search for the contract state saved in the transaction's leader_receipt of the previous accepted transaction.

# Why

# Testing done

- Tested the new feature in the UI.

# Decisions made

A contract state rollback for a leader appeal (appeal on undetermined state) is not needed because the transaction will overwrite the contract state immediately when the transaction goes to the accepted state.

# Checks

- [x] I have tested this code

# Reviewing tips

Run studio, check code changes

# Depends on
This PR contains code from PR #787 

# User facing release notes

The contract state is restored after a successful validator appeal.


backend/consensus/base.py
```diff
@@ -1436,6 +1455,13 @@ async def handle(self, context):
         # Set up the iterator for rotating through the involved validators
         context.iterator_rotation = rotate(involved_validators)
 
+        # Reset the contract snapshot for the transaction
+        if not context.transaction.appeal_undetermined:
+            context.transactions_processor.set_transaction_contract_snapshot(
```
cristiam86, line 1460: @kstroobants why is it necessary to reset the contract_snapshot?

backend/consensus/base.py
```diff
@@ -1824,51 +1864,54 @@ async def handle(self, context):
...
-
-            # Register contract if it is a new contract
-            if context.transaction.type == TransactionType.DEPLOY_CONTRACT:
-                new_contract = {
-                    "id": context.transaction.data["contract_address"],
-                    "data": {
-                        "state": {
-                            "accepted": leader_receipt.contract_state,
-                            "finalized": {},
+        # Do not deploy or update the contract if validator appeal failed
```
cristiam86, line 1870: @kstroobants is this the condition for appeal failing? seems to me that is for a transaction being appealed

backend/consensus/base.py
kstroobants: I believe these lines can be removed. My thought was that when the validator appeal is successful I want to start in a clean state but actually because of the rollback the contract state will be the same as when it was executed in the beginning so contract_snapshot will contain the same value and will be set in the ProposingState. So I agree.

backend/consensus/base.py
kstroobants: We do not want to update the contract state when the appeal failed. E.g. if you have a write method that appends something to state variable then it is added multiple times.

- The first time you come to the accepted state (happy path or from successful validator/leader appeal) then context.transaction.appealed is false and we update the contract.
- When we appeal the transaction and go to committing, revealing and accept states then context.transaction.appealed is still on true when we arrive in accepted state. It is set to false in the database in the accepted state but the variable context.transaction.appealed still contains true. Hence we know it is a failed validator appeal and we do not update the contract.

backend/consensus/base.py
kstroobants: I will add it to the rollback function because the rollback transactions need to be treated as 'new' again as they might have different outcomes (accepted vs undetermined).

backend/consensus/base.py
```diff
@@ -1508,11 +1534,24 @@ async def handle(self, context):
             context.transaction.to_address
         )
 
+        # Set the contract snapshot for the transaction
+        if context.transaction.contract_snapshot is None:
```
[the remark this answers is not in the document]
kstroobants: Moved to accepted state in 00c9c3e06c8e11115863f71b090ebe3f8efcc71b

backend/consensus/base.py
kstroobants: Moved to rollback function 00c9c3e06c8e11115863f71b090ebe3f8efcc71b

cristiam86 (approved):

PR #801, opened 2025-01-08
feat: consensus frontend get method should show accepted and finalized contract state

### LINKED ISSUE #788: Consensus: frontend get method shows accepted and finalized contract state
-> 1. might be the easiest. Both require a change in GenlayerJS (https://github.com/yeagerai/genlayer-js/blob/d941a8d6342c62916c50246269cf1cc1db471357/src/contracts/actions.ts#L25)

Two options:
1. Get method does two eth_calls, one for accepted and one for finalized.
   - Frontend does two calls and shows both return values.
   - eth_call contains a flag to choose accepted or finalized state. Put default on accepted so that the write method does not need a flag.
   - _SnapshotView class read method need to select the correct state key based on the given flag.
2. Get method does one eth_call getting the whole state dictionary.
   - Parsing/decoding needs to be updated in GenlayerJS.
   - Doubt: _SnapshotView class read method will give the whole dictionary but the write method call of the frontend uses the _SnapshotView class read method. So somehow the accepted state from the dictionary needs to be given to the genvm processing and not the whole dictionary.

kstroobants:
Fixes #788

# What

- Added for loop in frontend to do two gen_calls. When read method button is pressed it will output the accepted and the finalized contact state.
- Added the state status as a data member of the _SnapshotView class and use it in the storage_read and storage_write methods.
- Removed the encoded_state from the contractSnapshot class as it is redundant.
- Added gen_call endpoint.

# Why

- To show both states with the read method.

# Testing done

- Tested the new feature in the studio.
- Did an JSONRPC request to gen_call.

# Decisions made

Read method button performs two eth_calls (one for accepted and one for finalized). Another option was two buttons.

# Checks

- [x] I have tested this code

# Reviewing tips

Run in studio and view code changes.

# Depends on
PR https://github.com/yeagerai/genlayer-js/pull/72
PR #799 (based on the pr branch)

# Todo
genlayer-studio code should point to the genlayer-js version containing the update.

# User facing release notes

When the read method button is pressed, it will output the accepted and the finalized contact state.


kstroobants:
both tests fail because they depend on genlayer-js

backend/consensus/base.py
```diff
@@ -1233,7 +1232,7 @@ async def process_validator_appeal(
                     # Get the previous state of the contract
                     if context.transaction.contract_snapshot:
                         previous_contact_state = (
-                            context.transaction.contract_snapshot.encoded_state
+                            context.transaction.contract_snapshot.states["accepted"]
```
cristiam86, line 1235: @kstroobants when a validator appeal suceeds, shouldn't we takle the latest finalized state instead of the accepted one?

kstroobants: Imagine case 1:
- tx1 finalized
- tx2, tx3 accepted
The contract_snapshot saved in tx3 in the db contains finalized state of tx1 and accepted state of tx2.

Imagine case 2:
- tx1, tx2 finalized
- tx3 accepted
The contract_snapshot saved in tx3 in the db contains finalized state of tx2 and accepted state of tx2.

If tx3 validator appeal succeeds then we rollback tx3. We want for previous_contact_state, the state of tx2. For both cases, taking the accepted state will give what we want.

frontend/src/components/Simulator/ContractMethodItem.vue
```diff
@@ -47,19 +48,29 @@ const formatResponseIfNeeded = (response: string): string => {
...
-      const resultString = abi.calldata.toString(result);
-      responseMessage.value = formatResponseIfNeeded(resultString);
-    } else {
-      responseMessage.value = '<genlayer.client is undefined>';
+    const responseMessages = [
+      responseMessageAccepted,
+      responseMessageFinalized,
+    ];
+
+    for (let i = 0; i < 2; i++) {
```
cristiam86, line 61:
```suggestion
const responseMessages = [
      responseMessageAccepted,
      responseMessageFinalized,
    ] = await Promise.all([
    callReadMethod(props.name, unfoldArgsData(calldataArguments.value),'latest-nonfinal'),
    callReadMethod(props.name, unfoldArgsData(calldataArguments.value),'latest-final'),
    ])
```

cristiam86 (changes requested):

cristiam86 (approved):

kstroobants:
Better version in #1152

PR #820, opened 2025-01-16
feat: consensus leader rotation update

### LINKED ISSUE #664: Consensus: leader rotation update
So remove old leader, add 1 new validator, choose new leader.

Currently there is an iterator to rotate but the leader becomes a validator and is not removed

kstroobants:
Fixes #664 

# What

- Leader rotation is updated
   - Old leader removed
   - New validator added
   - Random leader selected from the validators
- Added config_rotation_rounds field to Transactions database table.
   - For now hardcoded to 3 rotations.
   - This should be populated by the frontend like LeaderOnly. I will create an issue.
- An error is thrown when there are no validators left when trying to add a new validator to the transaction.
- Removed the rotation iterator.

# Why

# Testing done

- Tested the new feature in the studio and with the consensus unit tests.

# Decisions made

- Interpreted number of rotation to be equal to the number of changing the leader. So if there is always a majority disagreement and 3 rotation rounds configured then there were 4 leaders.
- The seed to select the new validator is based on the time. Same function used when 5 validators are selected. Function and seed configuration is not changed. In the real implementation every transaction has a seed.

# Checks

- [x] I have tested this code

# Reviewing tips

Run in the studio. Check the code changes.

# Depends on
PR #657 

# User facing release notes

Rotation is of the leader is updated to remove the old leader, pick a new validator and select a new leader.


cristiam86 (approved):

kstroobants:
Merged into main by PR #851

PR #826, opened 2025-01-16
feat: raise JSONRPCError when VITE_IS_HOSTED is True for blocked endpoints

### LINKED ISSUE #818: feat(backend): use `VITE_IS_HOSTED` to block simulator endpoints
So that hosted versions can block attackers from things like creating/deleting validators

kstroobants:
Fixes #818 

# What

- Raise JSONRPCError when VITE_IS_HOSTED is True for blocked endpoints
- Added a wrapper
- Added a protection for appeals to not be able to set it on True in the wrong transaction state with a CURL call

# Why

- To protect the hosted version

# Testing done

- Tested the new feature with CURL command

# Decisions made

- Added a wrapper for all the methods listed in nginx/rpc-blocker.lua -> restricted_methods
- For every blocked method a call returns 
`{"error":{"code":-32000,"data":{},"message":"Non-allowed operation","name":"JSONRPCError"},"id":1,"jsonrpc":"2.0"}`
   - Set code to -32000
   - Set data to be empty as it does not matter because the method is blocked. If needed I can add them.

# Checks

- [x] I have tested this code

# Reviewing tips

Run the studio with VITE_IS_HOSTED on True. Check the code changes.
You can use the following command:
```
curl -X POST http://localhost:4000/api -H "Content-Type: application/json" -d '{
    "jsonrpc": "2.0",
    "method": "sim_setFinalityWindowTime",
    "params": [1000],
    "id": 1
}'
```

# User facing release notes

Block RPC methods also in the backend for hosted version

### DIFF
--- a/backend/database_handler/transactions_processor.py
+++ b/backend/database_handler/transactions_processor.py
@@ -346,7 +346,10 @@ class TransactionsProcessor:
         transaction = (
             self.session.query(Transactions).filter_by(hash=transaction_hash).one()
         )
-        transaction.appealed = appeal
+        if (transaction["status"] == TransactionStatus.ACCEPTED.value) or (
+            transaction["status"] == TransactionStatus.UNDETERMINED.value
+        ):
+            transaction.appealed = appeal
 
     def set_transaction_timestamp_accepted(
         self, transaction_hash: str, timestamp_accepted: int = None
--- a/backend/database_handler/transactions_processor.py
+++ b/backend/database_handler/transactions_processor.py
@@ -346,8 +346,8 @@ class TransactionsProcessor:
         transaction = (
             self.session.query(Transactions).filter_by(hash=transaction_hash).one()
         )
-        if (transaction["status"] == TransactionStatus.ACCEPTED.value) or (
-            transaction["status"] == TransactionStatus.UNDETERMINED.value
+        if (transaction.status == TransactionStatus.ACCEPTED.value) or (
+            transaction.status == TransactionStatus.UNDETERMINED.value
         ):
             transaction.appealed = appeal
 
--- a/backend/protocol_rpc/endpoints.py
+++ b/backend/protocol_rpc/endpoints.py
@@ -2,7 +2,7 @@
 import random
 import json
 import eth_utils
-from functools import partial
+from functools import partial, wraps
 from typing import Any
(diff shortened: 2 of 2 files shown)


backend/protocol_rpc/endpoints.py
```diff
@@ -52,6 +52,22 @@
 from flask import request
 from flask_jsonrpc.exceptions import JSONRPCError
 import base64
+import os
+
+
+####### WRAPPER TO BLOCK ENDPOINTS FOR HOSTED ENVIRONMENT #######
+def check_vite_hosted(func):
```
cristiam86, line 59: @kstroobants this is not a good name for this wrapper. It should be something similar to `check_forbidden_method_in_hosted_studio`

cristiam86 (changes requested):

cristiam86 (approved):

backend/database_handler/transactions_processor.py
```diff
@@ -346,7 +346,10 @@ def set_transaction_appeal(self, transaction_hash: str, appeal: bool):
         transaction = (
             self.session.query(Transactions).filter_by(hash=transaction_hash).one()
         )
-        transaction.appealed = appeal
+        if (transaction.status == TransactionStatus.ACCEPTED.value) or (
+            transaction.status == TransactionStatus.UNDETERMINED.value
+        ):
+            transaction.appealed = appeal
```
AgustinRamiroDiaz, line 352: I think this doesn't have to do with this PR

backend/protocol_rpc/endpoints.py
```diff
@@ -52,6 +52,22 @@
...
+    def wrapper(*args, **kwargs):
+        if os.getenv("VITE_IS_HOSTED") == "true":
+            raise JSONRPCError(
+                code=-32000,
+                message="Non-allowed operation",
+                data={},
+            )
+        return func(*args, **kwargs)
+
+    return wrapper
```
AgustinRamiroDiaz, line 70: I like the wrapper :D

backend/database_handler/transactions_processor.py
kstroobants: Yes and no. It gives protection when using the endpoint. Let me know if I should put it in another PR.

PR #841, opened 2025-01-22
fix: SIM-FE-Consensus appeal button should not stay green when not enough validators

### LINKED ISSUE #813: SIM-FE-Consensus appeal button should not stay green when not enough validators
Steps to reproduce:
1. Commented out code for appeal button in frontend/src/components/Simulator/TransactionItem.vue if commented.
2. Run studio.
3. Create no more than 5 validators.
4. Send a transaction (deploy new contract or write method).
5. Press the appeal button.
6. Appeal button stays green.
-> Backend put transaction.appealed on False but frontend is not aware.

kstroobants:
Fixes #813 

# What

- When transaction processor sets the transaction appealed field then a log event is sent.
- The log event is not shown in the terminal or in the logs UI because the RPC already says the transaction is appealed.
- The log event triggers a frontend transaction update to have the updated appealed field. This removes the need of setting appealed frontend variable to true when pressing the appeal button.

# Why

- To fix a bug

# Testing done

- Tested the bug fix in the UI

# Decisions made

- Emitting a log event is needed to update the frontend after the backend's database is updated. This removes the green color from the UI appeal button when there are no validators left to do the appeal.
- There is no need to send this log event every time the appeal function is called as throughout the consensus the status is also updated and send a log event to update the transaction in the frontend.

# Checks

- [x] I have tested this code

# Reviewing tips

Run UI and check code changes.

# Depends on
PR #657 

# User facing release notes

Appeal button color is updated correctly when no more validators a present to perform the appeal.


backend/consensus/base.py
```diff
@@ -1053,7 +1053,7 @@ async def process_validator_appeal(
             # When no validators are found, then the appeal failed
             print(e, transaction)
             context.transactions_processor.set_transaction_appeal(
-                context.transaction.hash, False
+                context.transaction.hash, False, self.msg_handler
```
cristiam86, line 1056: @kstroobants lets keep the msg_handler in the consensus. Call it after this method has returned.
Having the msg_handler into many components of the protocol would increase the coupling and will make it hard to improve/refactor/debug.

cristiam86 (changes requested):

backend/consensus/base.py
kstroobants: solved in commits below.

cristiam86 (approved):

PR #859, opened 2025-01-27
feat: update finality window value

### LINKED ISSUE #857: Update finality window value
The finality window value in the backend and the frontend should match when:
- Reloading the backend;
- Reloading the frontend;
- Sending an RPC request to 4000.

kstroobants:
Fixes #857

# What

- Frontend hot reload/ browser refresh: added webSocketClient.on connect to fetchFinalityWindowTime .
- Backend hot reload: added webSocketClient.on connect to fetchFinalityWindowTime. The connect already fetched the value so no need to send a logevent with set_finality_window_time. Hence, the init of consensusAlgorithm does not call set_finality_window_time.
- When whole system starts up: Added call to consensusStore.fetchFinalityWindowTime to frontend/src/hooks/useSetupStores.ts.
- RPC call: added log event when backend sets the finality window to update the frontend. And added webSocketClient.on finality_window_time_updated to fetchFinalityWindowTime.
- Added isLoading to fetchFinalityWindowTime to not display the value set by const. There is a delay until value is fetched from backend.

# Why

- To add more value to the user.

# Testing done

- Tested the new feature in the UI with VITE_PROXY_ENABLED on false and true.

# Decisions made

When the backend reloads it sets the finality window to the env variable, not what was in the frontend. So frontend adapts to backend.

# Checks

- [x] I have tested this code

# Reviewing tips

Check the code changes. Try in the UI.

# Depends on
PR #657 

# User facing release notes

Finality window time matches in frontend and backend when reloading.


cristiam86:
@kstroobants please fix backend unit tests

PR #867, opened 2025-01-31
feat: consensus deduct processing time appeal

### LINKED ISSUE #665: Consensus: deduct processing time appeal
When the appeal failed, the processing time of the appeal should not be part of the appeal window time

Low priority

kstroobants:
Fixes #665 

# What

- Added timestamp_appeal and appeal_processing_time to the transactions table in the database.
- appeal_processing_time contains the time it takes to process the appeals. This means that the processing time is equal to the difference of the time when the appeal is finished (so failed, because if success then the processing time is reset) and the time when the appeal was asked for (press the appeal button or RPC request). The time is increased for multiple appeals.
- The processing time is taken into account when checking if the transaction's waiting time is bigger than the finality window.

# Why

- To not count for the processing time of the appeal in the finality window.

# Testing done.

- Tested the new feature in the UI.
- Added a check to the consensus testcases.

# Decisions made

NA.

# Checks

- [x] I have tested this code

# Reviewing tips

- Check in the UI.
- Check the code changes.

# Depends on
PR contains code from PR #657.

# User facing release notes

The finality window does not count for the processing time of the appeals.


backend/database_handler/transactions_processor.py
```diff
@@ -297,11 +301,13 @@ def set_transaction_appeal(self, transaction_hash: str, appeal: bool):
         )
         # You can only appeal the transaction if it is in accepted or undetermined state
         # Setting it to false is always allowed
-        if (
-            (not appeal)
-            or (transaction.status == TransactionStatus.ACCEPTED)
-            or (transaction.status == TransactionStatus.UNDETERMINED)
-        ):
+        if appeal:
```
cristiam86, line 304:
```suggestion
        if appeal and ((transaction.status == TransactionStatus.ACCEPTED) or (
                transaction.status == TransactionStatus.UNDETERMINED
            ))
            self.set_transaction_timestamp_appeal(transaction, int(time.time()))
        transaction.appealed = appeal
```
Just avoid to repeat the code transaction.appealed = appeal

backend/database_handler/transactions_processor.py
```diff
@@ -400,3 +406,29 @@ def get_newer_transactions(self, transaction_hash: str):
...
+            )
+        transaction.timestamp_appeal = timestamp_appeal
+
+    def set_transaction_appeal_processing_time(
+        self, transaction_hash: str, appeal_processing_time: int | None = None
+    ):
+        transaction = (
+            self.session.query(Transactions).filter_by(hash=transaction_hash).one()
+        )
+        if appeal_processing_time == 0:
```
cristiam86, line 425: @kstroobants here also refine the if statement.
This: 
```
if appeal_processing_time == 0:  
    transaction.appeal_processing_time = 0
```    
Is the same as ```transaction.appeal_processing_time += appeal_processing_time```

cristiam86 (changes requested):
@kstroobants why shouldn't we count for the processing time of the appeal in the finality window? where is this specified?

kstroobants:
> @kstroobants why shouldn't we count for the processing time of the appeal in the finality window? where is this specified?

The specs is not up to date with this feature but it is valid. I confirmed with Kiril. I believe the reason is that validators don't have the power to influence the 30min finality window by processing the request slower.

backend/database_handler/transactions_processor.py
kstroobants: Would have been nice but the transaction.status check exists to prevent putting it on True when the transaction is in another state. So, I am not going to do this.

backend/database_handler/transactions_processor.py
kstroobants: This line exist to reset the time. I have split set_transaction_appeal_processing_time into two functions to make it more readable (set and reset) in dcd9deb6ebc66ac34f2b6a92464029345e98b358.

backend/database_handler/transactions_processor.py
cristiam86: alright, it is just that the nested ifs are very ugly and hard to read. 
What about this:
```python
if not appeal:
    transaction.appealed = appeal
    self.session.commit()
elif transaction.status in (TransactionStatus.ACCEPTED, TransactionStatus.UNDETERMINED):
    transaction.appealed = appeal
    self.set_transaction_timestamp_appeal(transaction, int(time.time()))

```

kstroobants: I like it! Solved in a2c95b5eb112528d7e051a20004fda2d028d475b

cristiam86 (approved):

cristiam86 (approved):

PR #891, opened 2025-02-10
fix: deploy event and tutorial error

### LINKED ISSUE #893: bug: tutorial and deployment issues in the studio
- deployment is not properly detected anymore
- tutorial is broken when there are zero validators
- deployment transaction is not persisted in the list on reload

denishacquin:
# What
Fixes #893 

- **Fixed deployment flow**; We were listening to a transaction event which was unreliable and moreover did not carry the contract address in its payload anymore, resulting in an error when subsequently fetching the contract abi. We are now listening to a dedicated **deployed_contract** event instead.
- **Fixed broken tutorial**: when 0 validators, some conditions in the templates were hiding the sections with contract constructor and methods, resulting in a broken tutorial. Added relevant adding exceptions in the UI to handle this edge case. Furthermore, the tutorial was broken because the mock data we were using didn't match the structure of the payloads of contracts anymore, so I update the mock data to reflect it.
- **Fixed transaction store persister race condition**: when deploying a contract, we first clear the existing transactions then add the new deploy transaction. The clear was happening after the adding, resulting in an empty tx list on reload. I made sure the order would be respected by awaiting the clear method call before adding the new transaction.

# Why
- to fix some breaking bugs

# Testing done
- tested the bug fixes

# Decisions made

- Decided to listen to **deployed_contract** event (instead of transaction status update) as a trigger to add deployed contracts in the client state.
- Added unit test for the new composable **useContractListener**.

# Checks

- [x] I have tested this code

# Reviewing tips

Try out tutorial with zero validators, and contract deployments.

# User facing release notes

- Fixed a bug with the deployment of contracts
- Fixed a bug with the tutorial tour
- Fixed a bug where deployment transactions were not persisted in state

### DIFF
--- a/frontend/src/components/Simulator/ContractInfo.vue
+++ b/frontend/src/components/Simulator/ContractInfo.vue
@@ -2,7 +2,7 @@
 import PageSection from '@/components/Simulator/PageSection.vue';
 import { CheckCircleIcon } from '@heroicons/vue/24/outline';
 import EmptyListPlaceholder from '@/components/Simulator/EmptyListPlaceholder.vue';
-import { useNodeStore } from '@/stores';
+import { useNodeStore, useUIStore } from '@/stores';
 import { useContractQueries, useShortAddress } from '@/hooks';
 import { UploadIcon } from 'lucide-vue-next';
 
@@ -15,6 +15,7 @@ defineProps<{
 
 const emit = defineEmits(['openDeployment']);
 const { isDeployed, address, contract } = useContractQueries();
+const uiStore = useUIStore();
 </script>
 
 <template>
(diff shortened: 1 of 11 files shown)


cristiam86:
> @denishacquin please solve the pre-commit check fail
> 
> When I clear my chrome browser data (see image below) and reload the page, then all the issues still exist. Screenshot 2025-02-11 at 16 56 11
> 
> When I press on the right top to show the tutorial again, then it looks good except that the gen_getContractSchemaForCode error is still there when the tutorial finishes(see images below). Screenshot 2025-02-11 at 16 54 31

This seems like a different error than what Denis fixed. Please confirm if this is due to some contracts stored locally but not in the backend, so when you request their deployment, it fails.

cristiam86 (approved):

denishacquin:
> > @denishacquin please solve the pre-commit check fail
> > When I clear my chrome browser data (see image below) and reload the page, then all the issues still exist. Screenshot 2025-02-11 at 16 56 11
> > When I press on the right top to show the tutorial again, then it looks good except that the gen_getContractSchemaForCode error is still there when the tutorial finishes(see images below). Screenshot 2025-02-11 at 16 54 31
> 
> This seems like a different error than what Denis fixed. Please confirm if this is due to some contracts stored locally but not in the backend, so when you request their deployment, it fails.

Indeed, had the same errors, cleared the local studio storage and that fixed it.

PR #904, opened 2025-02-12
feat: update genvm to 0.0.14

kp2pml30:
Fixes no issue

# What

- updated genvm version

# Why

- to update genvm version

# Testing done

- ran existing tests

# Decisions made

# Checks

- [x] I have tested this code

# Reviewing tips

# User facing release notes

## Breaking changes:
1. class annotation -> extending a class
autofix regexp: `perl -pe 'BEGIN{undef $/;} s/\@gl\.contract\n\s*class (\w*):/class $1(gl.Contract):/smg' -i`
2. `gl.message.*_account` was renamed to `gl.message.*_address`
autofix regexp: `perl -pe 's/\b(gl\.message\.(sender|origin|contract))_account\b/$1_address/g' -i`
3. if you use VecDB please ensure that `import numpy` happens before `from genlayer import *`
otherwise you will get following error: `ImportError: please import `numpy` before `from genlayer import *` if you wish to use VecDB`
4. I decided that all classes used within storage must be annotated. This way users will know that they work with something from "storage" and not a regular python object, for that reason `@allow_storage` annotation was introduced
5. multi file contract sources now need to be in <zip> `contract/*` instead of `src/*`

## Features:
1. balances were introduced:
	- `self.balance`
	- `ContractAt(...).balance`
	- `ContractAt(...).emit_transfer(value=...)`
	- `@gl.public.write.payable` decorator
	- method `def __handle_undefined_method__(self, method_name: str, args: list[typing.Any], kwargs: dict[str, typing.Any])`
	- method `def __receive__(self)`
2. eth module was reworked and became part of public api: https://sdk.genlayer.com/section_public_packages/genlayer.py.eth.html
3. if llm request fails with REQUEST_TIMEOUT/SERVICE_UNAVAILABLE/GATEWAY_TIMEOUT genvm will retry after a couple of seconds
4. added handling for case when validator reaches non-det block N while leaders maximum was N-1 (result will be ContractError("deterministic_violation"))
5. xai provider was verified (+ atoma got merged)

## Invisible changes:
- significant RAM usage optimization via mmap
- reworked some rust code to be async
- ported cpython wasm building to nix to facilitate:
	- hashes for every dependency
	- have a frozen world state
	- execute within sandbox/chroot for nearly-perfect isolation for deterministic builds

## QoL:
1. GenVM is now 100% statically linked (I am not 100% sure about macos build...). Which means that glibc incompatibilities won't happen again
2. I reworked error traces for reflection-related code a bit: now it uses exception notes instead of throwing dependent exception. Example:
```
TypeError: use `bigint` or one of sized integers please, see https://docs.genlayer.com/developers/intelligent-contracts/storage
during processing type `int`
during generating field `var: int`
during processing type Storage (declared at {'file': '/contract.py', 'line': 12})
```


frontend/src/components/Simulator/TransactionItem.vue
```diff
@@ -60,30 +65,52 @@ watch(
 );
 
 function prettifyTxData(x: any): any {
+  const oldResult = x?.consensus_data?.leader_receipt?.result;
```
kp2pml30, line 68: this ugly code tries to prettify values (calldata and results) for user in json viewer

cristiam86 (approved):

PR #907, opened 2025-02-13
 chore: Pull Docker Images Instead of Building by Default

YeagerAI-Bulat:
#### Pull Docker Images Instead of Building by Default

---

#### **Summary**
This PR modifies the behavior of the Docker setup by changing the default from building images locally to pulling them from a Docker Hub registry. This helps improve deployment speed, ensures consistency across environments, and reduces the complexity of managing local builds. Deployment time reduced from 15-19 minutes to ~8 minutes.

---

#### **Changes**
- Updated `docker-compose.yml`:
  - Added `image` fields for services that were previously built locally.
  - Introduced `pull_policy: always` for services to ensure they always fetch the latest image.
  - Adjusted `frontend` service to use `pull_policy: build`, allowing it to be rebuilt on each deployment.
- Updated `roles/genlayer/tasks/main.yml`:
  - Changed `build: always` to `build: policy` in `community.docker.docker_compose_v2` to align with the new behavior.

---

#### **Why This Change?**
- **Faster Deployments**: Pulling pre-built images from a registry eliminates the need to build locally every time.
- **Consistency**: Ensures that all instances use the same version of the images, preventing inconsistencies due to local environment differences.
- **Reduced Build Complexity**: Simplifies the setup for new developers and deployment environments by avoiding unnecessary local builds.

---

#### **Potential Impact**
- Developers who previously relied on local builds may need to manually trigger a rebuild if they are making changes to Docker images.
- If necessary, specific services can be overridden to allow local builds using Docker Compose overrides or by modifying `pull_policy`.

---

#### **Testing**
- Verified that the new configuration successfully pulls images instead of building them. Github actions run: https://github.com/yeagerai/genlayer-simulator-infra/actions/runs/13301335817
- Ensured that services start up correctly using the updated `docker-compose.yml`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -11,7 +11,13 @@ services:
       - ./traefik.yaml:/etc/traefik/traefik.yaml:ro
 
   frontend:
-    # image: yeagerai/simulator-frontend:latest
+    image: yeagerai/simulator-frontend:latest
+    build:
+      context: ./
+      dockerfile: ./docker/Dockerfile.frontend
+      target: ${FRONTEND_BUILD_TARGET:-final}
+      args:
+        - VITE_*
     pull_policy: build
     ports:
       - "${FRONTEND_PORT}:8080"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -110,7 +110,6 @@ services:
       - ./.env:/app/webrequest/.env
       - ./webrequest:/app/webrequest
     environment:
-    environment:
       - FLASK_SERVER_PORT=${WEBREQUESTPORT}
       - WEBREQUESTSELENIUMPORT=${WEBREQUESTSELENIUMPORT}
       # TODO: remove this in production
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 services:
   traefik:
     profiles: ["studio"]
-    image: traefik:v3.2
+    image: traefik:v3.3.3
     pull_policy: always
     ports:
       - "80:80"
@@ -12,13 +12,13 @@ services:
 
   frontend:
     image: yeagerai/simulator-frontend:latest
+    pull_policy: build
     build:
       context: ./
       dockerfile: ./docker/Dockerfile.frontend
       target: ${FRONTEND_BUILD_TARGET:-final}
       args:
         - VITE_*
-    pull_policy: build
     ports:
       - "${FRONTEND_PORT}:8080"
     volumes:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -1,7 +1,8 @@
 services:
   traefik:
     profiles: ["studio"]
-    image: traefik:v3.2
+    image: traefik:v3.3.3
+    pull_policy: always
```
AgustinRamiroDiaz, line 5: I don't think it's necessary to mention pull_policy always. When the image changes, docker should be intelligent enough to pull it again
```suggestion
```

docker-compose.yml
```diff
@@ -10,6 +11,8 @@ services:
       - ./traefik.yaml:/etc/traefik/traefik.yaml:ro
 
   frontend:
+    image: yeagerai/simulator-frontend:${STUDIO_DOCKER_IMAGE_VERSION:-latest}
+    pull_policy: build
     build:
```
AgustinRamiroDiaz, line 16: I think we should remove the `build` and `volumes` sections, since they are not needed anymore

AgustinRamiroDiaz (approved):
I'm approving to unblock, but please take a look at my comments

docker-compose.yml
```diff
@@ -43,6 +46,8 @@ services:
       traefik.http.routers.frontend.tls: true
 
   jsonrpc:
+    image: yeagerai/simulator-jsonrpc:${STUDIO_DOCKER_IMAGE_VERSION:-latest}
+    pull_policy: always
```
AgustinRamiroDiaz, line 50: are you sure this doesn't interfere with local development?

I'm pretty sure from past experiences that this breaks local development

docker-compose.yml
```diff
@@ -10,6 +11,8 @@ services:
       - ./traefik.yaml:/etc/traefik/traefik.yaml:ro
 
   frontend:
+    image: yeagerai/simulator-frontend:${STUDIO_DOCKER_IMAGE_VERSION:-latest}
+    pull_policy: build
```
AgustinRamiroDiaz, line 15: why is this one the only one with `build`?

AgustinRamiroDiaz (changes requested):
Sorry, I got confused. I thought this PR was on genlayer-simulator-infra.

I think these changes might interfere with local development. Why don't you copy the docker-compose.yml and put it on genlayer-simulator-infra? That way it won't interfere
Another idea: we could create a `docker-compose.prod.yml` in this repo

docker-compose.yml
YeagerAI-Bulat: https://github.com/yeagerai/genlayer-simulator-infra/issues/72#issuecomment-2626876413

cristiam86:
This is solved by the CLI. Cloning the Studio project locally and starting it with docker-compose is for developers

PR #922, opened 2025-02-18
feat: tests mock llm calls 

### LINKED ISSUE #915: tests: mock llm calls

kstroobants:
Fixes #915

# What

- Added mock plugin.
- Set response for contracts that use LLM calls.
- Modified integration tests GH workflow to always use mock unless it is a PR review approved.

# Why

- To not be dependent on external LLM calls in the tests.

# Testing done

- Tested the new feature by running the testcases with mock and checking the logs.
- Writing the env file variable inside the GH workflow with true or false works.
- I did not test if it uses openai llm call when the PR review is approved...

# Decisions made

- providers_schema.json is general for the mock, so having responses is not set as a requirement etc.
- You need to hardcode the response linked to a prompt. The prompt is parsed from the contract code. Another option is to put the responses in a list without the prompts and iterate over the list but you need to somehow know when the prompt is an equivalence prompt.
- I did not take into account responses for appeal rounds.
- All 5 validators use the same response.

# Checks

- [x] I have tested this code

# Reviewing tips

Check code changes. Try the testcases.

# User facing release notes

Mock LLM calls in integration tests.


tests/integration/conftest.py
```diff
@@ -22,6 +22,67 @@ def setup_validators():
...
 
 
+def setup_mock_validators(responses, comparison_result):
+    # First delete any existing mock providers
+    providers_response = post_request_localhost(
+        payload("sim_getProvidersAndModels")
+    ).json()
+    assert has_success_status(providers_response)
+
+    for provider in providers_response["result"]:
```
cristiam86, line 32: @kstroobants try to debug this

cristiam86 (changes requested):

kstroobants:
Better version in https://github.com/genlayerlabs/genlayer-studio/pull/1190

PR #951, opened 2025-02-28
fix: update o1 models with correct parameters

### LINKED ISSUE #947: fix: validators default config
OpenAI has changed some API parameters, max_tokens doesn't work for all the models.
For now, we can just remove this parameter.

kstroobants:
Fixes #947 

# What

- Updated the `openai_o1-mini.json` and `openai_o1-preview.json` configuration files to replace `max_tokens` with `max_completion_tokens` and set its value to 5000.
- Modified the `providers_schema.json` files in both backend and frontend to include a new `max_completion_tokens` field with appropriate constraints and comments. The comments are used to filter the fields for the presets.
- Updated the `ProviderModal.vue` component to handle the new configuration logic for OpenAI models. `max_tokens` and `temperature` is applied to GPT-4 models while only `max_completion_tokens` is applied to `o1` models.
- Refactored the `get_openai_stream` function in `llms.py` to dynamically include relevant parameters (`temperature`, `max_tokens`, `max_completion_tokens`) based on their presence in the configuration.

# Why

- To fix a bug related to incorrect parameter usage in OpenAI model configurations.
- To ensure that the application correctly handles different model configurations, particularly for the `o1` models which do not require a `temperature` field.

# Testing done

- Verified that the local studio correctly applies the new `max_completion_tokens` configuration for `o1` models.
- Tested the application to ensure that `temperature` is only applied to GPT-4 models.

# Decisions made

- Decided to remove the `temperature` field for `o1` models as it is fixed and not required.
- Chose to dynamically construct the parameters for the OpenAI API call to ensure flexibility and correctness.
- I have set an interval for `max_completion_tokens` but we might need to conduct experiments to determine a better interval.

# Checks

- [x] I have tested this code

# Reviewing tips

- Pay attention to the changes in the configuration logic in `ProviderModal.vue` to ensure it aligns with the new requirements.
- Review the refactoring in `llms.py` to confirm that the parameter handling is robust and flexible.
- Update your database with the new presets by removing them from the database with the CLI.

# User facing release notes

- Updated OpenAI model configurations to use `max_completion_tokens` and removed `temperature` for `o1` models.


denishacquin (approved):
LGTM

PR #956, opened 2025-03-03
feat: simulate transaction

### LINKED ISSUE #835: feat: simulate transaction
We should be able to call write method with the eth_call method and get its result without writing to the state

kstroobants:
Fixes #835

# What

- Updated the `Node` class to allow `transaction_hash` to be optional because there is no transaction hash when simulating the transaction.
- Added a new parameter `validators_registry` to the `gen_call` function to retrieve a validator.
- Implemented logic to simulate transactions in the `gen_call` function by checking if the method is write operation.
- Modified the `useContractQueries` hook to make `blockId` optional.

# Why

- To enable transaction simulation, which allows users to test transactions without executing them on the blockchain.

# Testing done

- Added a new button and response message handling for simulating transactions in the `ContractMethodItem.vue` component. Remove it if not needed before merging.
- Tested the transaction simulation feature to ensure it correctly simulates write operations.
- Verified that the contract state is not updated when simulating transactions.

# Decisions made

- Decided to use the first available validator for transaction simulation to simplify the implementation.
- Chose to make `transaction_hash` and `blockId` optional to enhance flexibility.
- Changed the mock node to a real validator node because the read method could potentially also contain an LLM call.
- to_address is required in gen_call specification so this means that the contract is already deployed. How would we do a gen_call for deployment?
- To check if the method is a read or write I go over all methods of the contract. A calldata.split("method") could work but between method and the actual method name I see different chars depending on the call:
   - DecodedMethodCallData(calldata=b'\x0e\x06method\\get_storage’)
   - DecodedMethodCallData(calldata=b'\x16\x04args\r\x1411\x06methodtupdate_storage')
   - DecodedMethodCallData(calldata=b'\x0e\x06methodlreset_storage')

# Checks

- [x] I have tested this code

# Reviewing tips

- Focus on the changes in the `gen_call` function and the new UI elements in `ContractMethodItem.vue`.
- Verify that the optional parameters are handled correctly in the code.

# Depends on
Contains code from PR #801.

# User facing release notes

Added the ability to simulate transactions, allowing users to test contract interactions without executing them on the blockchain.


cristiam86 (approved):

cristiam86:
@kstroobants please solve conflicts

kstroobants:
Better version in #1150.

Issue #933, opened 2025-02-24
feat: show GenVM and Studio version in frontend


cristiam86:
The GenVM version should be returned from the gen_protocolVersion endpoint.
The Studio version should be pulled form a local VERSION file in the root of the project

PR #961, opened 2025-03-05
feat: return genvm and studio versions from gen_protocolVersion endpoint

Henadek:
Fixes #issue-number-here

# What

- added a gen_protocolVersion rpc endpoint that reads version of genvm and studio from backend docker image and frontend package config
- modified dockerfile to capture the versions during build in a shared_data volume

# Why

- to allow easy view and tracking of what versions of genvm and studio are used 

# Testing done

- rpc integration test to verify the both versions of genvm and studio exists for every build of the software

# Decisions made

- Having a shared_data volume allows the backend rpc endpoint capture the version of the systems in use 

# Checks

- [ ] I have tested this code

# Reviewing tips

# User facing release notes

### DIFF
--- a/docker/Dockerfile.frontend
+++ b/docker/Dockerfile.frontend
@@ -8,19 +8,23 @@ COPY ./examples src/assets/examples
 COPY ./.env ./.env
 COPY ../backend/node/create_nodes/providers_schema.json /app/src/assets/schemas/providers_schema.json
 
+RUN apk add --no-cache jq \
+    && mkdir -p /app/shared /app/tmp \
+    && chmod -R 777 /app/shared /app/tmp \
+    && jq -r '.version' package.json > /app/tmp/studio_version.txt
 
 FROM base AS dev
-ENTRYPOINT ["npm", "run", "dev"]
+ENTRYPOINT sh -c "cp -r /app/tmp/studio_version.txt /app/shared/studio_version.txt && rm -rf /app/tmp && npm run dev"
 
 FROM base AS builder
 RUN npm run build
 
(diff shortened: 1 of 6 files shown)


Henadek:
#933

docker/Dockerfile.frontend
```diff
@@ -8,19 +8,23 @@ COPY ./examples src/assets/examples
 COPY ./.env ./.env
 COPY ../backend/node/create_nodes/providers_schema.json /app/src/assets/schemas/providers_schema.json
 
+RUN apk add --no-cache jq \
+    && mkdir -p /app/shared /app/tmp \
+    && chmod -R 777 /app/shared /app/tmp \
+    && jq -r '.version' package.json > /app/tmp/studio_version.txt
```
kstroobants, line 14: ./frontend/package.json contains:
```
package"name": "genlayer-simulator",
"version": "0.0.0",
```
Will this version be updated in the future with the correct version seen in github?

kstroobants: Created issue #990

Henadek: @kstroobants  There is this 
`package"name": "genlayer-simulator",
"version": "0.0.0",
`
and there is also this

`  "dependencies": {
    ...
    "genlayer-js": "^0.7.0",
`
I will have to verify which of these versions is required, or if the "genlayer-simulator" version will always the current  "genlayer-js" version

kstroobants:
@cristiam86 For me this PR is blocked by issue #990. If we cannot update that version number then it does not make sense to get the version from `frontend/package.json`?

Issue #895, opened 2025-02-10
feat: import contract by address


cristiam86:
As a user, I would like to provide a deployed contract address and interact with it.

In the files tab, we should add a third button to "Import" a contract in the top right section.

![Image](https://uploads.linear.app/2becad37-331a-45e1-ab14-372ad0cad3ab/f7c19cb9-3480-4953-a440-eee5bdcb0af7/4aff12e4-dcb7-4c05-b159-afaccc98f427?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzJiZWNhZDM3LTMzMWEtNDVlMS1hYjE0LTM3MmFkMGNhZDNhYi9mN2MxOWNiOS0zNDgwLTQ5NTMtYTQ0MC1lZWU1YmRjYjBhZjcvNGFmZjEyZTQtZGNiNy00YzA1LWIxNTktYWZhY2NjOThmNDI3IiwiaWF0IjoxNzQ0MDMwNTgyLCJleHAiOjMzMzE0NTkwNTgyfQ.aGHEpdHWBAgjPxvUHKpSSkxI3UnurWJVs76G87-RoLo)

This button should ask for an address. After entering the address, we should pull the contract code from the backend and create a new contract in the frontend with that code.

After importing the contract, users should be able to interact with the contract deployed at that address

PR #977, opened 2025-03-10
feat: added rpc endpoint to fetch existing deployed contract from add…

Henadek:
…ress #895

Fixes #issue-number-here

# What

- made endpoint to fetch contract data from a contract address
- made a db session context manager to allow necessary parts query db

# Why

- to make it possible to interact with an existing deployed contract
- contracts do not have to be re-created for the same use case, they can just be imported and used as is

# Testing done

- rpc integration test to verify the contract data is fetched from an address

# Decisions made
- having an enpoint to fetch the contract data makes it easy for the ui to interact with the backend

# Checks

- [ ] I have tested this code

# Reviewing tips

# User facing release notes

### DIFF
--- a/backend/protocol_rpc/endpoints.py
+++ b/backend/protocol_rpc/endpoints.py
@@ -13,7 +13,7 @@ import backend.node.genvm.origin.calldata as genvm_calldata
 from backend.database_handler.contract_snapshot import ContractSnapshot
 from backend.database_handler.llm_providers import LLMProviderRegistry
 from backend.rollup.consensus_service import ConsensusService
-from backend.database_handler.models import Base
+from backend.database_handler.models import Base, Transactions, db_session
 from backend.domain.types import LLMProvider, Validator, TransactionType
 from backend.node.create_nodes.providers import (
     get_default_provider_for,
@@ -726,6 +726,29 @@ def get_contract(consensus_service: ConsensusService, contract_name: str) -> dic
     }
 
 
+def get_contract_by_address(address: str) -> dict[str:Transactions] | None:
+    import base64
+
+    with db_session() as session:
+        if AccountsManager(session).is_valid_address(address):
+            deployed_contract: Transactions = (
+                session.query(Transactions)
+                .filter(
+                    Transactions.type == TransactionType.DEPLOY_CONTRACT,
+                    Transactions.to_address == address,
+                )
+                .one()
+            )
+            if deployed_contract:
+                return {
+                    "transactions": deployed_contract.hash,
+                    "contract_code": base64.b64decode(
+                        deployed_contract.data.get("contract_code")
+                    ),
+                    "contract_address": deployed_contract.to_address,
+                }
+
+
 def register_all_rpc_endpoints(
     jsonrpc: JSONRPC,
     msg_handler: MessageHandler,
@@ -894,4 +917,8 @@ def register_all_rpc_endpoints(
     register_rpc_endpoint(
         partial(get_block_by_hash, transactions_processor),
         method_name="eth_getBlockByHash",
+    )
+    register_rpc_endpoint(
+        get_contract_by_address,
+        method_name="gen_getContractByAddress",
     )
--- a/tests/integration/deploy/test_deploy.py
+++ b/tests/integration/deploy/test_deploy.py
@@ -2,12 +2,13 @@ from pathlib import Path
 import zipfile
 import io
 import base64
-
+import pytest
 from tests.common.request import (
     deploy_intelligent_contract,
     write_intelligent_contract,
     payload,
     post_request_localhost,
+    get_contract_by_address,
 )
 
 from tests.common.response import (
@@ -45,3 +46,12 @@ def test_deploy(setup_validators, from_account):
     res = call_contract_method(contract_address, from_account, "test", [])
 
     assert res == "123"
+
+
+@pytest.mark.asyncio
+async def test_get_contract_by_address():
+    status_code, contract = await get_contract_by_address("test_address")
+    result = contract["result"]
+
+    assert status_code == 200
+    assert isinstance(result, (dict | None))
--- a/backend/database_handler/models.py
+++ b/backend/database_handler/models.py
@@ -1,5 +1,4 @@
 from typing import List, Optional
-
 from sqlalchemy import (
     BigInteger,
     Boolean,
@@ -14,6 +13,7 @@ from sqlalchemy import (
     text,
     ForeignKey,
     Text,
+    create_engine,
 )
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm import (
@@ -22,9 +22,44 @@ from sqlalchemy.orm import (
     mapped_column,
     MappedAsDataclass,
     relationship,
+    sessionmaker,
+    scoped_session,
 )
+from contextlib import contextmanager
+from os import environ
 import datetime
 import enum
+
+
+def get_db_name(database: str) -> str:
+    return "genlayer_state" if database == "genlayer" else database
+
+
+database_name_seed = "genlayer"
+db_uri = f"postgresql+psycopg2://{environ.get('DBUSER')}:{environ.get('DBPASSWORD')}@{environ.get('DBHOST')}/{get_db_name(database_name_seed)}"
+engine = create_engine(db_uri, echo=True, pool_size=50, max_overflow=50)
+
+# setup scoped session
+SessionFactory = sessionmaker(bind=engine)
+Session = scoped_session(SessionFactory)
+
+
+@contextmanager
+def db_session():
+    """
+    Context manager for managing database sessions.
+
+    Usage:
+    with db_session() as session:
+        # Use the session object for database operations
+        session.query(...)
+    """
+    session = Session()
+    try:
+        yield session
+    finally:
+        session.close()
+        Session.remove()
 
 
 class TransactionStatus(enum.Enum):
--- a/tests/common/request.py
+++ b/tests/common/request.py
@@ -276,3 +276,10 @@ def decode_nested_data(data):
         return [decode_nested_data(item) for item in data]
     else:
         return data
+
+
+async def get_contract_by_address(address: str):
+    payload_data = payload("gen_getContractByAddress", address)
+    raw_response = post_request_localhost(payload_data)
+    status_code, parsed_raw_response = raw_response.status_code, raw_response.json()
+    return status_code, parsed_raw_response


backend/protocol_rpc/endpoints.py
```diff
@@ -726,6 +726,29 @@ def get_contract(consensus_service: ConsensusService, contract_name: str) -> dic
     }
 
 
+def get_contract_by_address(address: str) -> dict[str:Transactions] | None:
+    import base64
+
+    with db_session() as session:
```
kstroobants, line 732: Is it possible to use the same approach as in this code below. The session already exists and is created in server.py? If yes, then remove the added functions in backend/database_handler/models.py:
```
register_rpc_endpoint(
        partial(clear_db_tables, request_session),
        method_name="sim_clearDbTables",
    )
```

Henadek:
comments are resolved

backend/protocol_rpc/endpoints.py
```diff
@@ -726,6 +726,29 @@ def get_contract(consensus_service: ConsensusService, contract_name: str) -> dic
...
+            deployed_contract: Transactions = (
+                session.query(Transactions)
+                .filter(
+                    Transactions.type == TransactionType.DEPLOY_CONTRACT,
+                    Transactions.to_address == address,
+                )
+                .one()
+            )
+            if deployed_contract:
+                return {
```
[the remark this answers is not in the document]
Henadek: unused attribs, removed

backend/protocol_rpc/endpoints.py
```diff
@@ -726,6 +726,29 @@ def get_contract(consensus_service: ConsensusService, contract_name: str) -> dic
     }
 
 
+def get_contract_by_address(address: str) -> dict[str:Transactions] | None:
+    import base64
+
+    with db_session() as session:
+        if AccountsManager(session).is_valid_address(address):
+            deployed_contract: Transactions = (
+                session.query(Transactions)
```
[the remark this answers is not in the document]
Henadek: I didn't know the CurrentState table can be used to achieve same as the Transactions at least in this case. Switched to CurrentState now.

tests/integration/deploy/test_deploy.py
```diff
@@ -45,3 +46,12 @@ def test_deploy(setup_validators, from_account):
     res = call_contract_method(contract_address, from_account, "test", [])
 
     assert res == "123"
+
+
+@pytest.mark.asyncio
+async def test_get_contract_by_address():
```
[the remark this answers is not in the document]
Henadek: done

frontend/src/views/Simulator/ContractsView.vue
```diff
@@ -6,6 +6,61 @@ import { v4 as uuidv4 } from 'uuid';
...
+const showInputModal = ref(false);
+const contractAddress = ref(''); // Store the input field value
+
+function toggleInputModal() {
+  showInputModal.value = !showInputModal.value;
+}
+
+async function getContractByAddress(address: string) {
+  try {
+    const request: JsonRPCRequest = {
```
kstroobants, line 24: Can you do it with JsonRpcService like all the other rpc requests? Have a look in the ts files on how other requests are implemented.

Henadek: done

kstroobants:
@cristiam86 When the contract is imported from an address. Should Run and Debug frontend tab show the deployed contract with read and write methods or should it should the button to deploy the contract? I guess the first.
This is how it look now (it was a storage deployed contract):
Screenshot 2025-03-25 at 12 00 33

kstroobants:
@Henadek Can you make the UX design the same as the whole studio theme?
- So buttons, text field, title, background window, text font etc.
- The window title is gone in the dark mode.

Screenshot 2025-03-25 at 11 54 54
Screenshot 2025-03-25 at 11 54 44

frontend/src/views/Simulator/ContractsView.vue
```diff
@@ -5,7 +5,51 @@ import { ref } from 'vue';
 import { v4 as uuidv4 } from 'uuid';
 import ContractItem from '@/components/Simulator/ContractItem.vue';
 import MainTitle from '@/components/Simulator/MainTitle.vue';
-import { useEventTracking } from '@/hooks';
+import { useRpcClient, useEventTracking } from '@/hooks';
+import { InboxArrowDownIcon } from '@heroicons/vue/24/solid';
```
kstroobants, line 9: Could you use an icon from https://lucide.dev/icons/ to have the same style like the other two buttons (line 3)

tests/integration/deploy/test_deploy.py
```diff
@@ -45,3 +47,29 @@ def test_deploy(setup_validators, from_account):
...
+        ),
+    ],
+)
+@pytest.mark.asyncio
+async def test_get_contract_by_address(address: str, status: int):
+    status_code, contract = await get_contract_by_address(address)
+    result = contract.get("result")
+
+    assert status_code == status
+    if result is None:
```
kstroobants, line 71: @cristiam86 question regarding testing vs shipping fast: Both parametrized testcases go this if statement. Should we also have one for the else statement (so deploying a contract and getting it back)?

Henadek:
> @Henadek Can you make the UX design the same as the whole studio theme?
> 
> * So buttons, text field, title, background window, text font etc.
> * The window title is gone in the dark mode.
> 
> Screenshot 2025-03-25 at 11 54 54 Screenshot 2025-03-25 at 11 54 44

now aligned theme with studio theme

frontend/src/views/Simulator/ContractsView.vue
Henadek: done

backend/protocol_rpc/endpoints.py
```diff
@@ -726,6 +726,21 @@ def get_contract(consensus_service: ConsensusService, contract_name: str) -> dic
     }
 
 
+def get_contract_by_address(
+    request_session: Session, accounts_manager: AccountsManager, address: str
+) -> dict[str:CurrentState] | None:
+    if accounts_manager.is_valid_address(address):
```
cristiam86, line 732: @Henadek @kstroobants there is a missing error return here in case it is not a valid address.

backend/protocol_rpc/endpoints.py
```diff
@@ -726,6 +726,21 @@ def get_contract(consensus_service: ConsensusService, contract_name: str) -> dic
...
+        deployed_contract: CurrentState = (
+            request_session.query(CurrentState)
```
cristiam86, line 734: @Henadek @kstroobants we shouldn't be doing queries here. That's the responsibility of the inner services. Endpoints are for validations and execution of services

tests/integration/deploy/test_deploy.py
cristiam86: This is not a valid unit test, there should be one for each case. Unit tests shouldn't have any logic flows

cristiam86 (changes requested):

backend/protocol_rpc/endpoints.py
Henadek: resolved

backend/protocol_rpc/endpoints.py
Henadek: raising an exception 
`raise AccountNotFoundError(address, f"Account {address} does not exist.")`
and returning None when address is valid but not found.

Is this okay?

frontend/src/views/Simulator/ContractsView.vue
```diff
@@ -1,11 +1,54 @@
...
-import { FilePlus2, Upload } from 'lucide-vue-next';
+import { FilePlus2, Upload, ArrowDownToLine } from 'lucide-vue-next';
 import { ref } from 'vue';
 import { v4 as uuidv4 } from 'uuid';
 import ContractItem from '@/components/Simulator/ContractItem.vue';
 import MainTitle from '@/components/Simulator/MainTitle.vue';
-import { useEventTracking } from '@/hooks';
+import { useRpcClient, useEventTracking } from '@/hooks';
+
+const rpcClient = useRpcClient();
```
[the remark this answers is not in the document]
Henadek: I think other endpoints are called in the same way. 
Examples:
https://github.com/yeagerai/genlayer-studio/blob/5f6f8bc61df35cc322a7bef3fd28e9ba4d1a7cad/frontend/src/stores/transactions.ts#L16

https://github.com/yeagerai/genlayer-studio/blob/5f6f8bc61df35cc322a7bef3fd28e9ba4d1a7cad/frontend/src/stores/transactions.ts#L90

https://github.com/yeagerai/genlayer-studio/blob/5f6f8bc61df35cc322a7bef3fd28e9ba4d1a7cad/frontend/src/stores/consensus.ts#L18

backend/protocol_rpc/endpoints.py
cristiam86: ok

frontend/src/views/Simulator/ContractsView.vue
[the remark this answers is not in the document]
Henadek: https://github.com/yeagerai/genlayer-studio/pull/977/commits/a8f157ebe002027849ce262304cf21ffef9d0a28
https://github.com/yeagerai/genlayer-studio/pull/977/commits/3a0b6446e93c4a02f7ed5a1ef341254141dfb970
these committs attempts to fix

frontend/src/views/Simulator/ContractsView.vue
```diff
@@ -1,12 +1,55 @@
...
 
+const { rpcClient } = useRpcDependencyStore();
+const showInputModal = ref(false);
+const contractAddress = ref(''); // Store the input field value
+
+function toggleInputModal() {
+  showInputModal.value = !showInputModal.value;
+}
+
+async function getContractByAddress(address: string) {
```
cristiam86, line 18: @Henadek I feel like this should be a method in the hook useContractQueries and the contractsStore that can be used for Vue components directly without too much logic.
The RPC methods are valid, but we shouldn't be calling the RPC from view components, we need to go through hooks and stores

cristiam86 (changes requested):

frontend/src/views/Simulator/ContractsView.vue
Henadek: This change is effected. Rpc method is moved into contractStore.
https://github.com/yeagerai/genlayer-studio/pull/977/commits/09bc925e183941dce4d00358eebf879fe5a80020

tests/integration/deploy/test_deploy.py
```diff
@@ -45,3 +48,34 @@ def test_deploy(setup_validators, from_account):
...
+async def test_get_contract_by_address_invalid():
+    address = "test_address"
+    status_code, _ = await get_contract_by_address(address)
+
+    assert status_code == 400
+
+
+@pytest.mark.asyncio
+async def test_get_contract_by_address_not_exist():
+    address = "0x9C778c9688dAA91FDa539399B817C8732c284F18"
```
cristiam86, line 63: @Henadek how do you know that this address exists? I think you should mock the database or something so you can test this properly

cristiam86 (changes requested):

tests/integration/deploy/test_deploy.py
Henadek: now mocked the rpc call

PR #994, opened 2025-03-11
feat: store validators private key on their db record

### LINKED ISSUE #991: feat: Store validators private key on their DB record

kstroobants:
Fixes #991

# What

- Added functionality to store the validator's private key in their database record.
- Updated the `Validator` class and related database models to include the `private_key` field.
- Modified the `create_validator` and `create_random_validators` functions to handle the private key.

# Why

- To store a private key of each validator so they can sign transactions.

# Testing done

- Verified that the private key is correctly stored in the database when creating a validator.
- Cloning a validator in the UI creates a new validator with a different private key.

# Decisions made

- Decided to store the private key as a nullable string in the database to accommodate existing records without keys.

# Checks

- [x] I have tested this code

# Reviewing tips

- Focus on the changes.

# User facing release notes

- Validators have their private keys now stored in their database records.


backend/consensus/base.py
```diff
@@ -1157,9 +1158,12 @@ async def process_validator_appeal(
                     )
 
                     # Get the previous state of the contract
-                    previous_contact_state = (
-                        context.transaction.contract_snapshot.encoded_state
-                    )
+                    if context.transaction.contract_snapshot:
```
cristiam86, line 1161: @kstroobants what is this change about?

kstroobants: the consensus unit test was failing because set_transaction_contract_snapshot() in test_helpers.py does not contain an implementation so the contract snapshot is kept on None.
Adapted in 03e06f4.

PR #1005, opened 2025-03-17
feat: add Geth to Hardhat Service for Data Persistence

mpaya5:
### **What**  
- **Integrated Geth into the Hardhat service/container**  
  - Allows for persistent storage of blockchain data, including contracts, transactions, and accounts.  
  - Ensures that data is not lost upon container restarts.  
  - Improves local development experience by maintaining state across runs.  

### **Why**  
- **Persistence**: Previously, Hardhat’s in-memory blockchain would reset on each restart.  
- **Improved Testing & Debugging**: Keeps contract deployments and transaction history intact.  
- **Better Local Development**: Avoids re-deploying contracts and losing chain state.  

### **Testing Done**  
- Verified that Hardhat + Geth runs correctly in the container.  
- Ensured blockchain state persists across container restarts.  
- Deployed contracts and checked they remain available after service restart.  

### **User Facing Release Notes**  
- **Added Geth to the Hardhat service**: Ensures blockchain data persists across restarts.  
- **No more losing contracts and transactions**: Develop and test with a persistent blockchain state.


mpaya5:
It's not ready to be merged, we need to implement the `unlock` of the accounts in the AccountManager.

PR #1036, opened 2025-03-28
feat: adjust value in current_state based on transaction values

### LINKED ISSUE #1028: feat: adjust value in current_state based on transaction values

kstroobants:
Fixes #1028

# What

- Implemented value transfer between accounts in the consensus mechanism when the transaction is in the `ACCEPTED` state.
- Added error handling for transactions when the value sent is bigger than the sender's balance.
- Updated GenVM call to support a transaction value parameter.
- Added automatic account funding when creating new accounts in the frontend.
- Updated frontend components to handle account creation asynchronously.
- Frontend shows the value input field for write method when the decorator of the write method in the intelligent contract is `@gl.public.write.payable`.

# Todo
- Hardhat log error when sending value=10:
```
hardhat-1             |   Sender doesn't have enough funds to send tx. The max upfront cost is: 10 and the sender's balance is: 0.
jsonrpc-1             | [CONSENSUS_SERVICE]: Error forwarding transaction: {'code': -32000, 'message': "Sender doesn't have enough funds to send tx. The max upfront cost is: 10 and the sender's balance is: 0.", 'data': {'message': "Sender doesn't have enough funds to send tx. The max upfront cost is: 10 and the sender's balance is: 0.", 'data': None}}
```
- Showing the value field depends on the payable key in the contract schema. This schema will be updated in the next GenVM release. Once it is released:
   - Check if the code works.
   - Remove the frontend/src/types/contract.ts file and put that code in genlayer-js.

# Why

- To adjust the balaces of the accounts in the database based on the value send with the transaction. 

# Testing done

- Tested the funding of a newly create frontend account.
- Tested balance transfer when sending a value lower and higher than the sender's balance.
- Tested sending value when the decorator is not payable. This results in a GenVM error.

# Decisions made

- Decided to fund new frontend accounts with 10000. By calling `sim_fundAccount` the frontend account address is stored in the backend database and can be used when sending transactions with a value.

# Checks

- [x] I have tested this code

# Depends on
This PR contains code from PR #1027.

# Reviewing tips

- Focus on the changes in the consensus layer and GenVM for value handling.
- Review the error handling logic in the transactions processor and RPC endpoints.
- Check the frontend changes for asynchronous account creation and funding.

# User facing release notes

- Transaction value is handled by GenVM and stored in the database.
- New frontend accounts are automatically funded upon creation.

### DIFF
--- a/backend/consensus/base.py
+++ b/backend/consensus/base.py
@@ -2045,6 +2045,24 @@ class AcceptedState(TransactionState):
                     context, leader_receipt.pending_transactions, "accepted"
                 )
 
+                # Update the balance of the sender account
+                from_balance = context.accounts_manager.get_account_balance(
+                    context.transaction.from_address
+                )
+                context.accounts_manager.update_account_balance(
+                    context.transaction.from_address,
+                    from_balance - context.transaction.value,
+                )
+
+                # Update the balance of the recipient account
+                to_balance = context.accounts_manager.get_account_balance(
+                    context.transaction.to_address
+                )
+                context.accounts_manager.update_account_balance(
+                    context.transaction.to_address,
+                    to_balance + context.transaction.value,
+                )
+
         else:
             context.transaction.appealed = False
 
@@ -2242,13 +2260,30 @@ def _emit_transactions(
             data = {
                 "calldata": pending_transaction.calldata,
             }
-        context.transactions_processor.insert_transaction(
-            context.transaction.to_address,  # new calls are done by the contract
-            pending_transaction.address,
-            data,
-            value=0,  # we only handle EOA transfers at the moment, so no value gets transferred
-            type=transaction_type.value,
-            nonce=nonce,
-            leader_only=context.transaction.leader_only,  # Cascade
-            triggered_by_hash=context.transaction.hash,
-        )
+
+        try:
+            context.transactions_processor.insert_transaction(
+                context.transaction.to_address,  # new calls are done by the contract
+                pending_transaction.address,
+                data,
+                value=0,  # we only handle EOA transfers at the moment, so no value gets transferred
+                type=transaction_type.value,
+                nonce=nonce,
+                leader_only=context.transaction.leader_only,  # Cascade
+                triggered_by_hash=context.transaction.hash,
+                accounts_manager=context.accounts_manager,
+            )
+        except ValueError as e:
+            context.msg_handler.send_message(
+                LogEvent(
+                    "consensus_event",
+                    EventType.ERROR,
+                    EventScope.CONSENSUS,
+                    str(e),
+                    {
+                        "transaction_hash": context.transaction.hash,
+                        "sender_address": context.transaction.from_address,
+                    },
+                    transaction_hash=context.transaction.hash,
+                )
+            )
--- a/backend/database_handler/transactions_processor.py
+++ b/backend/database_handler/transactions_processor.py
@@ -17,6 +17,7 @@ from backend.database_handler.contract_snapshot import ContractSnapshot
 import os
 from sqlalchemy.orm.attributes import flag_modified
 from backend.domain.types import MAX_ROTATIONS
+from backend.database_handler.accounts_manager import AccountsManager
 
 from backend.rollup.consensus_service import ConsensusService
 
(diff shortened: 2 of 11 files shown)


kstroobants:
@cristiam86 I have some doubts:
1. Hardhat log error when sending value>0. 
```
hardhat-1             |   Sender doesn't have enough funds to send tx. The max upfront cost is: 10 and the sender's balance is: 0.
jsonrpc-1             | [CONSENSUS_SERVICE]: Error forwarding transaction: {'code': -32000, 'message': "Sender doesn't have enough funds to send tx. The max upfront cost is: 10 and the sender's balance is: 0.", 'data': {'message': "Sender doesn't have enough funds to send tx. The max upfront cost is: 10 and the sender's balance is: 0.", 'data': None}}
```
With the addTransaction we send a value to the consensusMain contract. So should we send in genlayer-js this value parameter at the genlayer transaction level instead of at the rollup transaction level?

2. Should the hardhat account balance be the same as the backend database account balance? I believe no because the hardhat is the rollup balance and the backend database the genlayer balance. But on the other hand, we have an eth_getBalance endpoint which is an eth method and gets the balance from the backend database. If yes, we can add this code when calling `sim_fundAccount`:
```
web3.provider.make_request("hardhat_setBalance", [
    address,
    hex(balance * 10**18)  # needs to be in hex format, balance ETH
])
```

3. Is the value of 10000 enough when init a frontend account? Adapt balance with https://linear.app/genlayer-labs/issue/DXP-429/frontend-manage-account-balance

4. When I have metamask wallet connected and I set a value for a write method then I cannot send it. How can we have GEN in the wallet?
Screenshot 2025-06-28 at 19 01 04

kstroobants:
> @cristiam86 I have some doubts:
> 
> 1. Hardhat log error when sending value>0.
> 
> ```
> hardhat-1             |   Sender doesn't have enough funds to send tx. The max upfront cost is: 10 and the sender's balance is: 0.
> jsonrpc-1             | [CONSENSUS_SERVICE]: Error forwarding transaction: {'code': -32000, 'message': "Sender doesn't have enough funds to send tx. The max upfront cost is: 10 and the sender's balance is: 0.", 'data': {'message': "Sender doesn't have enough funds to send tx. The max upfront cost is: 10 and the sender's balance is: 0.", 'data': None}}
> ```
> 
> With the addTransaction we send a value to the consensusMain contract. So should we send in genlayer-js this value parameter at the genlayer transaction level instead of at the rollup transaction level?
> 
> 2. Should the hardhat account balance be the same as the backend database account balance? I believe no because the hardhat is the rollup balance and the backend database the genlayer balance. But on the other hand, we have an eth_getBalance endpoint which is an eth method and gets the balance from the backend database. If yes, we can add this code when calling `sim_fundAccount`:
> 
> ```
> web3.provider.make_request("hardhat_setBalance", [
>     address,
>     hex(balance * 10**18)  # needs to be in hex format, balance ETH
> ])
> ```
> 
> 3. Is the value of 10000 enough when init a frontend account? Adapt balance with https://linear.app/genlayer-labs/issue/DXP-429/frontend-manage-account-balance
> 4. When I have metamask wallet connected and I set a value for a write method then I cannot send it. How can we have GEN in the wallet?
> 
> Screenshot 2025-06-28 at 19 01 04

Regarding (4) metamask calls eth_getBalance which calls our backend database current_state table. 

### New plan:
Hardhat enabled:
- hardhat EOA -insert-transaction-> consensusMain -on-acceptance-> ghostcontract
- backend EOA value decreases when insert_transaction and contract value increases when in accepted.

Hardhat disabled:
- backend EOA value decreases when insert_transaction and contract value increases when in accepted.

We do not have hardhat ghostcontracts but I will use the genlayer contract address to simplify it

Basically balances in backend db and hardhat should be the same. Maybe I add a row for consensusMain address in backend db.
- We need to fund hardhat accounts when calling sim_fundAccount to solve error in (1).
- Call sim_fundAccount when adding metamask account to frontend to solve error in (4). There is an issue when I first open the metamask extension (it calls eth_getBalance which returns 0), then add metamask account to frontend and increase balance in backend db. Sending a tx still fails because metamask thinks the balance is still 0 (refreshing the browser does not work)...

@cristiam86 Hardhat balance is in ETH and backend db balance is in GEN, can this work?

cristiam86:
> ```
> 
> ```
> 
> 
>     
> 
>   
> 
> With the addTransaction we send a value to the consensusMain contract.

Summary of the discussion: we store balances in the studio DB to avoid overcomplexity with hardhat (which is also optional to run)

cristiam86:
started on a newer PR

PR #1041, opened 2025-03-31
fix: successful validator appeal has no contract_snapshot_supplier

kstroobants:
Fixes #DXP-174

# What

- Reset the contract snapshot for the current transaction when a rollback occurs in the `ConsensusAlgorithm` class.
- Added a call to `set_transaction_contract_snapshot` with `None` to clear the snapshot for the transaction.

# Why

- To fix a bug where the contract snapshot was not being reset during a transaction rollback. When a validator appeal is successful then in the `AcceptedState` it needs the contract_snapshot_supplier. But it was not set in the `ProposingState` as there was still a contract_snapshot in the transaction.

# Testing done

- Tested a successful validator appeal in the UI.

# Decisions made

- Decided to reset the contract_snapshot to `None` to ensure a clean state for the transaction. We could keep it as it will be same when the transaction is processed again but then we need more logic in the `AcceptedState`'s handle function.

# Checks

- [x] I have tested this code

# Reviewing tips

- Focus on the changes made in the `ConsensusAlgorithm` class, specifically around the transaction rollback logic.
- Ensure that the reset logic does not interfere with other transaction processing steps.

# User facing release notes

- Fixed an issue where the contract snapshot of the appealed transaction was not reset during transaction rollbacks.


cristiam86 (approved):
I understand that this fixes the bug, but I do think that we need to re-think all the contract snapshots management since it is getting to hard to follow

kstroobants:
> @kstroobants I understand that this fixes the bug, but I do think that we need to re-think all the contract snapshots management since it is getting to hard to follow. First, lets separate what are actions to update the contract state and keep the contract snapshot as read-only

I think about the re-think.
About the read-only, I have this open PR #946

PR #1080, opened 2025-04-09
feat: trigger newtransaction for contract dependent transactions in consensus

kstroobants:
Fixes #DXP-190

# What

- Integrated `_addTransaction` inside the Solidity contract when calling `emitTransactionAccepted` or `emitTransactionFinalized`.
- Introduced a new `internalMessageData` struct in the Solidity contract to handle internal message data send from Python.
- Added logic to get the transaction hash from the new transactions triggered by the internal messages.

# Why

- To create new transactions in the rollup when internal messages are sent.

# Testing done

- Tested the new feature by simulating transaction emissions and verifying the correct handling of transaction events.
- Verified the Solidity contract's ability to process internal messages and call `_addTransaction`.
- This was checked with the use of `tests/integration/contract_examples/test_multi_tenant_storage.py` and genlayer-explorer.

# Decisions made

- The `emitTransactionAccepted` and `emitTransactionFinalized` trigger the `newTransaction` events inside the solidity code so they have all the same rollup sender.
   - A pro is that we do not need to send for every message a call to _addTransaction from Python.
   - A con is that `consensus/base.py` is less readable by moving the `emitTransactionAccepted` and `emitTransactionFinalized` calls inside the `_emit_transactions`instead of keeping them inside the `AcceptedState` and `FinalizingState`.
- For deployment messages like in `tests/integration/deploy/test_deploy.py`, I kept the creation/calculation of the new contract address in python. I tried following the idea we implemented in the endpoints file to get the contract address back from the solidity but then somehow genvm used not the same contract address. I did not find where genvm sets this.

# Checks

- [x] I have tested this code

# Reviewing tips

- Focus on the changes in the `_emit_transactions` function and the handling of transaction events.
- Review the Solidity contract changes, particularly the integration of `_addTransaction` and the handling of `internalMessageData`.

# User facing release notes

- IC to IC transactions are now integrated in hardhat.


mpaya5 (approved):

backend/consensus/base.py
```diff
@@ -2159,12 +2153,22 @@ async def handle(self, context):
...
+                    context,
+                    leader_receipt.pending_transactions,
+                    "accepted",
+                    context.consensus_data.leader_receipt.node_config,
                 )
 
         else:
             context.transaction.appealed = False
 
+            context.consensus_service.emit_transaction_event(
```
cristiam86, line 2165: @kstroobants why is this only emitted when the transaction was appealed?

backend/consensus/base.py
```diff
@@ -2297,27 +2301,32 @@ async def handle(self, context):
...
             # Insert pending transactions generated by contract-to-contract calls
             _emit_transactions(
                 context,
                 context.transaction.consensus_data.leader_receipt.pending_transactions,
                 "finalized",
+                leader_receipt.node_config,
+            )
+        else:
+            # Send events in rollup to communicate the transaction is finalized
+            context.consensus_service.emit_transaction_event(
```
cristiam86, line 2314: @kstroobants why is this event being emitted if the tx status in undetermined?

cristiam86 (changes requested):
@kstroobants  in the case of acceptance or finalization, instead of emitting the emitTransactionFinalized and emitTransactionAccepted events inside the _emit_transactions method I would wait for that method to end and emit the Finalized or Accepted event on their corresponding Consensus State.

backend/consensus/base.py
kstroobants: when the appeal window is over then the tx goes from undetermined to finalized state. Shouldn't we finalize the transaction by emitting a finalized event?
(this code line is in the FinalizingState class)

kstroobants:
> @kstroobants in the case of acceptance or finalization, instead of emitting the emitTransactionFinalized and emitTransactionAccepted events inside the _emit_transactions method I would wait for that method to end and emit the Finalized or Accepted event on their corresponding Consensus State.

Do you mean like this 77f1c0922912a40d4268a444260e3bce6ffa9fb8?

cristiam86:
> > @kstroobants in the case of acceptance or finalization, instead of emitting the emitTransactionFinalized and emitTransactionAccepted events inside the _emit_transactions method I would wait for that method to end and emit the Finalized or Accepted event on their corresponding Consensus State.
> 
> Do you mean like this [77f1c09](https://github.com/yeagerai/genlayer-studio/commit/77f1c0922912a40d4268a444260e3bce6ffa9fb8)?

Yes, this is better. I would also call the consensus service to emit the events inside the _emit_messages function but lets address that in the future if we re-think the transaction states

cristiam86 (approved):

PR #1117, opened 2025-04-22
feat: SIM-FE Finalized status badge color

### LINKED ISSUE #924: SIM-FE Finalized status badge color
Feedback from weekly sync to not show it always on green..

Henadek:
…s ERROR

Fixes #924 

# What

- added condition to check for execution status if ERROR, set TransactionStatusBadge to Red regardless of if transaction is Finalized.

# Why

- to make it easy to detect when execution has an ERROR but still Finalizes

# Testing done

- manual test of the new feature for when execution_result equals ERROR, then Finalized tx should be in red

# Decisions made

# Checks

- [x] I have tested this code

# Reviewing tips

# User facing release notes

### DIFF
--- a/frontend/src/components/Simulator/TransactionItem.vue
+++ b/frontend/src/components/Simulator/TransactionItem.vue
@@ -227,7 +227,11 @@ function prettifyTxData(x: any): any {
       <TransactionStatusBadge
         :class="[
           'px-[4px] py-[1px] text-[9px]',
-          transaction.status === 'FINALIZED' ? '!bg-green-500' : '',
+          leaderReceipt.execution_result === 'ERROR'
+            ? '!bg-red-500'
+            : transaction.status === 'FINALIZED'
+              ? '!bg-green-500'
+              : '',
         ]"
       >
         {{ transaction.status }}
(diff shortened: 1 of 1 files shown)


kstroobants:
@Henadek Do you also see a performance degradation for the frontend?
- Main branch: when tx is inserted then I see it going from pending -> activated -> proposing -> ... -> finalized.
- This PR branch: when tx is inserted then I see it only appearing from commiting to finalized.
I see this when I add prompt = +"a" at the end of the ask_for_coin function of the wizard of coin contract (placement of plus sign is wrong so executing it gives an error). Here an llm call is made when self.have_coin is True so it goes slowly through the states. Still I see this for all the contracts but if there is no llm call everything goes very fast.

You wrote in the PR description that you did a manual test, did you see the same behaviour? If yes, please fix. I believe it is related to execution_result not being accessible in the not shown states because no leader voted yet and accessing it might give an error (check frontend console via inspect in your browser).

The issue was only stating to change the colour to red in the finalized state but I see that you also do it in other states. It is better so we already know something is weird.

kstroobants (changes requested):
Do you also see a performance degradation for the frontend?

- Main branch: when tx is inserted then I see it going from pending -> activated -> proposing -> ... -> finalized.
- This PR branch: when tx is inserted then I see it only appearing from commiting to finalized.
I see this when I add prompt = +"a" at the end of the ask_for_coin function of the wizard of coin contract (placement of plus sign is wrong so executing it gives an error). Here an llm call is made when self.have_coin is True so it goes slowly through the states. Still I see this for all the contracts but if there is no llm call everything goes very fast.

Did you see the same behaviour in your manual test? If yes, please fix. I believe it is related to execution_result not being accessible in the not shown states because no leader voted yet and accessing it might give an error (check frontend console via inspect in your browser).

The issue was only stating to change the colour to red in the finalized state but I see that you also do it in other states. It is better so we know something is weird earlier. 

Can you also check how it behaves when we appeal a transaction (the colour should be the default colour again until there is an error)? More specifically, validator appeal success and a leader appeal. Write me if you do not know how to check it.

kstroobants:
Solved with https://github.com/genlayerlabs/genlayer-studio/pull/1270

PR #1120, opened 2025-04-23
fix: generate contracts address when consensus_service is not running

cristiam86:
Fixes #DXP-258

# What
- Changed the handling of binary calldata to use base64 encoding instead of UTF-8 decoding
- Modified transaction hash handling to use provided hash from receipt or generate a new one if not available

# Why
- To properly handle binary data in contract messages that may contain non-UTF-8 characters
- To ensure transaction hashes are always available, even when not provided in the receipt
- To prevent potential decoding errors when processing contract messages

# Testing done
- Verified that binary calldata is properly encoded as base64
- Confirmed that transaction hashes are correctly assigned from receipt or generated when needed
- Tested contract message processing with various data types

### DIFF
--- a/backend/consensus/base.py
+++ b/backend/consensus/base.py
@@ -2433,12 +2433,9 @@ def _emit_messages(
     receipt: dict,
 ):
     for i, insert_transaction_data in enumerate(insert_transactions_data):
-        if receipt and "tx_ids_hex" in receipt:
-            transaction_hash = receipt["tx_ids_hex"][i]
-        else:
-            new_account = context.accounts_manager.create_new_account()
-            transaction_hash = new_account.address
-
+        transaction_hash = (
+            receipt["tx_ids_hex"][i] if receipt and "tx_ids_hex" in receipt else None
+        )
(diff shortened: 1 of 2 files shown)


backend/consensus/base.py
```diff
@@ -2429,6 +2433,12 @@ def _emit_messages(
     receipt: dict,
 ):
     for i, insert_transaction_data in enumerate(insert_transactions_data):
+        if receipt and "tx_ids_hex" in receipt:
+            transaction_hash = receipt["tx_ids_hex"][i]
+        else:
+            new_account = context.accounts_manager.create_new_account()
```
kstroobants, line 2439: Why do we want the transaction_hash in the CurrentState database table?
Use instead (option 2 preferred):
1. Account.create()
2. Pass none to transaction_hash of transactions_processor.insert_transaction(...). The default is None so no need to write it explicitly. This will generate a tx hash with _generate_transaction_hash().

kstroobants (changes requested):
fix comment

kstroobants (approved):

PR #1134, opened 2025-05-02
fix: consensus history updated before logevent

kstroobants:
Fixes #DXP-76

# What

- Added a new parameter `update_current_status_changes` to the `dispatch_transaction_status_update` method in `ConsensusAlgorithm` to control whether the status should be updated in the consensus history.
- Updated the `update_transaction_status` method in `TransactionsProcessor` to handle the new parameter and conditionally update the `current_status_changes`.
- Modified the `update_consensus_history` method to include an optional `extra_status_change` parameter for additional status changes.
- Simplified the Dockerfile by removing unnecessary steps for ARM64 platform.

# Why

- To provide more control over the transaction status update process, allowing for conditional updates to the current status changes.
- To streamline the Dockerfile and remove redundant steps, improving build efficiency.

# Testing done

- Tested in the studio sending a failed leader appeal.

# Decisions made

- Decided to add the `update_current_status_changes` parameter to provide flexibility in transaction status updates.

# Checks

- [x] I have tested this code

# Reviewing tips

- Focus on the changes in the `ConsensusAlgorithm` and `TransactionsProcessor` classes to understand the new parameter's impact.

# User facing release notes

- Introduced a new parameter to control transaction status updates.

### DIFF
--- a/backend/consensus/base.py
+++ b/backend/consensus/base.py
@@ -622,6 +622,7 @@ class ConsensusAlgorithm:
         transaction_hash: str,
         new_status: TransactionStatus,
         msg_handler: MessageHandler,
+        update_current_status_changes: bool = True,
     ):
         """
         Dispatch a transaction status update.
@@ -633,7 +634,11 @@ class ConsensusAlgorithm:
             msg_handler (MessageHandler): Handler for messaging.
         """
         # Update the transaction status in the transactions processor
-        transactions_processor.update_transaction_status(transaction_hash, new_status)
+        transactions_processor.update_transaction_status(
+            transaction_hash,
+            new_status,
+            update_current_status_changes,
+        )
 
         # Send a message indicating the transaction status update
         msg_handler.send_message(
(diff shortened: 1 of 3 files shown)


cristiam86:
@kstroobants please explain better why do we need this extra parameter, it is not clear from the auto-generated description :)

kstroobants:
> explain better why do we need this extra parameter, it is not clear from the auto-generated description

The issue is the timing and visibility of the consensus history.

## Original Problem:
When a transaction goes to UNDETERMINED state, two operations happen:
1. Status update via dispatch_transaction_status_update which send a logevent and triggers the frontend to get the transaction data;
2. Consensus history update via update_consensus_history.

The frontend was not showing the latest consensus round because the history update happened after the trigger.

## Solution:
If you switch the order of 1 and 2, then the frontend contains the consensus round but it is not showing the UNDETERMINED state in the current_status_changes (the arrowed list above the tables in the frontend) because the dispatch_transaction_status_update comes after. So solution is to add the UNDETERMINED already and not set it in dispatch_transaction_status_update.

## Why the parameter helps:
The update_current_status_changes and extra_status_change parameter allows us to control updating the state earlier in the history and preventing a duplication. This is important because we first record the status change as part of the consensus history update, then dispatch the transaction status without creating a duplicate entry in the history.

cristiam86 (approved):

PR #1212, opened 2025-06-18
feat(genvm): update to v0.1.3

kp2pml30:
Upgrade that enhances lua scripting capabilities

Note: `gltest --contracts-dir . tests/integration/icontracts/tests/` passed locally

## Changes
In process I also updated syntax of contracts a bit

In case you need old stdlib (i.e in dkol) you can use following versions (@cristiam86):
```json
{
  "py-genlayer": "1j12s63yfjpva9ik2xgnffgrs6v44y1f52jvj9w7xvdn7qckd379",
  "py-genlayer-multi": "0xn02qjjqlpxadhs124cwrr5ddqs7dxcc6m5a7c6801hv54l66p0",
  "py-lib-genlayer-embeddings": "09h0i209wrzh4xzq86f79c60x0ifs7xcjwl53ysrnw06i54ddxyi"
}
```

I also added `# v0.1.0` to top of most contracts. Right now it is not mandatory, but after some time to enable events users will be obligated to specify version which introduces events

## Features
1. now web module also has `.lua` script (studio uses default one)
2. now there are more functions exposed to lua, check out default code for [web](https://github.com/genlayerlabs/genvm/blob/v0.1.2/executor/modules/implementation/scripting/web-default.lua) and [llm](https://github.com/genlayerlabs/genvm/blob/v0.1.2/executor/modules/implementation/scripting/llm-default.lua)

## Dev tip

If you clone genvm repo, you can add following to `settings.json`:
```json
  "python.analysis.extraPaths": [
    "/home/kp2pml30/work/genlayer/genvm/runners/genlayer-py-std/src",
    "/home/kp2pml30/work/genlayer/genvm/runners/genlayer-py-std/src-emb"
  ],
  "Lua.workspace.library": [
    "/home/kp2pml30/work/genlayer/genvm/executor/modules/implementation/scripting/"
  ],
```

This way you will get typehints for both python and lua

## Summary by CodeRabbit

* **New Features**
  * Added recommended extension for Lua development in the workspace settings.

* **Bug Fixes**
  * Improved error handling and reporting in backend modules, including updated result code classifications and mappings.
  * Enhanced error handling and overload detection in LLM backend prompt execution.
  * Adjusted timeout error signaling in consensus and origin host modules.

* **Refactor**
  * Updated contract and integration test examples to use new API methods for contract access and prompt execution.
  * Standardized dependency annotations and version comments across contract examples and tests.
  * Unified interface for LLM prompt execution with context-based parameters.
  * Modified equality principle prompt access and nondeterministic prompt execution in contracts.

* **Chores**
  * Updated Docker backend to use a newer genvm release and improved setup steps.
  * Added version annotations and clarified dependency metadata in contract and test files.


backend/node/base.py
```diff
@@ -343,8 +343,8 @@ async def _run_genvm(
                 stderr="",
                 genvm_log=[],
                 result=genvmbase.ExecutionError(
-                    message="Execution timed out",
-                    kind=genvmbase.ResultCode.CONTRACT_ERROR,
+                    message="timeout",
```
kp2pml30: I will add this "timeout" to autogenerated constants in next release, but it is the correct message, as all VMErrors (previously misleadingly named ContractErrors start with snake case error code)

backend/node/genvm/base.py
```diff
@@ -302,11 +299,11 @@ async def consume_result(
     ) -> None:
         if type == ResultCode.RETURN:
             self._result = ExecutionReturn(ret=bytes(data))
-        elif type == ResultCode.ROLLBACK:
```
kp2pml30, line 305: I decided to rename this result codes to prevent future confusion. Name `USER_ERROR` stands for error that contract produced explicitly. But name `contract_error` would bring a lot of confusion right now due to a clash with previous naming

backend/validators/greyboxing.lua
```diff
@@ -13,79 +14,86 @@ function get_mock_response_from_table(table, message)
...
-				-- Return the matching response to gl.eq_principle_prompt_comparative request which contains a principle key in the payload
-				result = get_mock_response_from_table(args.host_data.mock_response.eq_principle_prompt_comparative, prompt.user_message)
-			elseif args.payload.output then
-				-- Return the matching response to gl.eq_principle_prompt_non_comparative request which contains an output key in the payload
-				result = get_mock_response_from_table(args.host_data.mock_response.eq_principle_prompt_non_comparative, prompt.user_message)
-			else
-				-- Return the matching response to gl.exec_prompt request which does not contain any specific key in the payload
-				result = get_mock_response_from_table(args.host_data.mock_response.response, prompt.user_message)
-			end
+	if ctx.host_data.mock_response then
```
kp2pml30, line 22: @kstroobants Is there a code that tests this part? I am not 100% sure I patched this mocks correctly

examples/contracts/football_prediction_market.py
```diff
@@ -75,11 +76,13 @@ def get_match_result() -> str:
 your output must be only JSON without any formatting prefix or suffix.
 This result should be perfectly parsable by a JSON parser without errors.
             """
-            result = gl.exec_prompt(task).replace("```json", "").replace("```", "")
+            result = (
+                gl.nondet.exec_prompt(task).replace("```json", "").replace("```", "")
+            )
             print(result)
-            return json.dumps(json.loads(result), sort_keys=True)
+            return json.loads(result)
```
kp2pml30, line 83: @cristiam86 here is the fix that avoids dumps... I remember you wanted to introduce it

kp2pml30:
I will fix pre-commit after CI passes

backend/validators/greyboxing.lua
kp2pml30: found it myself...
```
TEST_WITH_MOCK_LLMS=true gltest --contracts-dir . tests/integration/icontracts/tests/
```

backend/node/base.py
kstroobants: Do you mean that timeout will get a different ResultCode number?

I made a PR in the studio to remove this empty receipt when res is None because it is old dead code. I introduced this to catch a timeout error that I generated in a different way than we do now. 

At the moment, when genvm execution returns timeout info, I check it by `(receipt.result[0] == ResultCode.CONTRACT_ERROR) and (receipt.result[1:] == b"timeout")`

kstroobants (approved):

backend/node/base.py
kp2pml30: I mean that `message="Execution timed out"` does not align with what genvm produces for timeout (what you wrote for check is correct, though CONTRACT_ERROR was renamed to VM_ERROR (underlying `int` constant remains the same)) or VM_ERROR in general

cristiam86 (approved):

PR #1221, opened 2025-06-24
feat: custom plugin

kstroobants:
Fixes #DXP-410

# What

- Updated the `providers_schema.json` to include a new "custom" plugin option.
- Added support for custom plugin in `backend/validators/__init__.py` and `backend/validators/greyboxing.lua`.
Screenshot 2025-06-24 at 18 04 45
Screenshot 2025-06-24 at 21 20 17
Screenshot 2025-06-24 at 21 19 55
Screenshot 2025-06-24 at 21 21 39
Screenshot 2025-06-24 at 21 21 53

# Why

To support a custom plugin, allowing more flexibility for users to integrate their APIs.

# Testing done

- Verified that the new "custom" plugin option is correctly recognized and processed.
- When I use the hosted huggingface models, I get a lot of connection errors during the execution of the llm calls. But I believe my code works because with the custom plugin using openai gpt-4 it works perfectly.

# Decisions made

- Decided to redirect the llm call to a web call when using a custom plugin to access the API. I asked Kira how a real validator would do it and if it would make sense to implement it in genvm instead of the studio. She told me: "A real node uses the same modules and all built-in providers are written in rust. If a real node wants to use something custom or extend greyboxing or whatever, they are free to implement whatever they want in lua. There are two customization points: config or lua."
- Decided to send all the validator data via `host_data` because the `greyboxing.lua` script does not have access to the `config, plugin_config (key and url)`.

# Checks

- [x] I have tested this code

# Reviewing tips

Focus on the changes.

# Depends on
This PR contains code from #1212

# User facing release notes

Added support for custom plugins, allowing users to integrate their own solutions.

## Summary by CodeRabbit

* **New Features**
  * Added support for a new "custom" plugin type, enabling integration with external APIs via user-defined settings.
  * Users can configure API keys and endpoints for custom plugins within provider settings.
  * Custom plugin supports adjustable parameters like temperature and max tokens with defined constraints.
  * Validator execution now supports calling custom plugin APIs with proper request handling and response processing.
  * Added asynchronous availability checks for custom plugins to verify connectivity and credentials.

* **Bug Fixes**
  * Enhanced validation and error handling for custom plugin configurations.

* **Documentation**
  * Updated provider schema documentation to include new "custom" plugin options and constraints.

### DIFF
--- a/backend/consensus/base.py
+++ b/backend/consensus/base.py
@@ -53,6 +53,7 @@ from backend.database_handler.validators_registry import ValidatorsRegistry
 from backend.node.genvm.origin.result_codes import ResultCode
 from backend.consensus.types import ConsensusResult, ConsensusRound
 from backend.consensus.utils import determine_consensus_from_votes
+from backend.node.genvm.origin.base_host import get_code_slot
 
 type NodeFactory = Callable[
     [
@@ -2083,18 +2084,7 @@ class RevealingState(TransactionState):
 
         # Determine the consensus result
         votes_list = list(context.votes.values())
-        agree_count = votes_list.count(Vote.AGREE.value)
-        disagree_count = votes_list.count(Vote.DISAGREE.value)
-        timeout_count = votes_list.count(Vote.TIMEOUT.value)
-
-        if timeout_count > agree_count and timeout_count > disagree_count:
-            consensus_result = ConsensusResult.TIMEOUT
-        elif agree_count > disagree_count and agree_count > timeout_count:
-            consensus_result = ConsensusResult.MAJORITY_AGREE
-        elif disagree_count > agree_count and disagree_count > timeout_count:
-            consensus_result = ConsensusResult.MAJORITY_DISAGREE
-        else:
-            consensus_result = ConsensusResult.NO_MAJORITY
+        consensus_result = determine_consensus_from_votes(votes_list)
 
         # Send event in rollup to communicate the votes are revealed
         if len(context.consensus_data.leader_receipt) == 1:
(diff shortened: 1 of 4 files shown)


kstroobants:
@cristiam86: I do not know if we should merge this in main or if we should use the diff to describe in the docs how to add a custom plugin. Some user might have their model name in the url and not in the payload etc.

Also, I don't know how much this will be used because all hosted models I found are compatible with `openai-compatible, ollama, anthropic or google`. Do you know how difficult it is to add this compatibility to one of the four mentioned for a model that a user has locally or on their own servers?

cristiam86:
> @cristiam86: I do not know if we should merge this in main or if we should use the diff to describe in the docs how to add a custom plugin. Some user might have their model name in the url and not in the payload etc.
> 
> Also, I don't know how much this will be used because all hosted models I found are compatible with `openai-compatible, ollama, anthropic or google`. Do you know how difficult it is to add this compatibility to one of the four mentioned for a model that a user has locally or on their own servers?

we proceed with this, and add documentation on how it works and how developers can add their custom plugins

danieljrc888:
@kstroobants  Is it ok that it doesnt show any model after configuring a new provider preset ?

image


image

kstroobants:
> @kstroobants Is it ok that it doesnt show any model after configuring a new provider preset ?
> 
> image image

Solved in bcda031c7e5ff0e5ae93294fcfcb74c8d2053377

kstroobants:
> * When the model is unavailable (for example gpt-8o), I was able to create the provider preset and execute a transaction, but it was stuck in the PROPOSING state.

Yes, this was expected. I had hardcoded the availability to True because genvm does not have custom providers to check if the model works. I added the same check that genvm does for custom models.

Check adds yellow triangle to show there is something wrong:
Screenshot 2025-07-23 at 14 39 25

Model that is not working is not selectable as a validator:
Screenshot 2025-07-23 at 14 39 48

> 
> Suggested future improvement:
> 
> * Allow the user to test their provider preset in the creation modal, it can be a button.

We will unblock these stuck transaction and the logs should give the user more information what is wrong with their model.

danielrc888 (approved):

PR #1267, opened 2025-07-23
chore(deps): update genlayer test version

danielrc888:
Fixes #issue-number-here

# What

- changed thing a for b
- also did this other unrelated thing in my path

# Why

- to fix a bug

# Testing done

- tested the new feature
- tested the bug fix

# Decisions made

# Checks

- [ ] I have tested this code

# Reviewing tips

# User facing release notes


cristiam86 (commented):
@danielrc888 two comments here:
1. We should keep all the tools versions under 1.0.0 until mainnet. I see that GenLayer-test is at 2.2.0. We need to remove this version, go back to 0.x.y, and temporary change the release pipeline to avoid changing major versions.
2. For better test performance, we should always wait for transaction accepted status unless we want to test an appeal

danieljrc888:
> 2\. For better test performance, we should always wait for transaction acc



> @danielrc888 two comments here:
> 
> 1. We should keep all the tools versions under 1.0.0 until mainnet. I see that GenLayer-test is at 2.2.0. We need to remove this version, go back to 0.x.y, and temporary change the release pipeline to avoid changing major versions.
> 2. For better test performance, we should always wait for transaction accepted status unless we want to test an appeal

@cristiam86 

1. Yes, you're right. I've created this issue https://linear.app/genlayer-labs/issue/DXP-539/rollback-genlayer-test-version-to-0xy to make this rollback (Once this is done, we should immediately update genlayer-test on studio)
2. Latest genlayer-test version now waits for transaction accepted status by default

danieljrc888:
> why do we use --default-wait-retries 140 instead of 40?

@kstroobants in latest genlayer-test version --default-wait-interval was decreased from 10000 ms to 3000 ms, so to compensate overall time I've updated  --default-wait-retries to 140

PR #1269, opened 2025-07-27
feat: create local profile for db

kstroobants:
Fixes #DXP-515

# What

- Updated the `.env.example` file to include `local` in the `COMPOSE_PROFILES` setting, allowing for a local development environment alongside `hardhat`.
- Modified the `docker-compose.yml` to add a `local` profile for the `postgres` service, enabling for the use of an optional database.
- Changed the volume mapping for `postgres` to use a named volume `postgres_data` for persistent data storage or cleaning when using `-v`.

# Why

- To enhance the development setup by allowing a local profile for the database and to make it possible to clean it.

# Testing done

# Decisions made

# Checks

- [x] I have tested this code

# Reviewing tips

# User facing release notes

Added support for a local development profile in the Docker setup, allowing for easier database management and persistence.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -132,6 +132,7 @@ services:
         max-file: "3"
 
   postgres:
+    profiles: ["local"]
     image: postgres:16-alpine
     ports:
       - "${DBPORT}:5432"
@@ -152,9 +153,9 @@ services:
       options:
         max-size: "10m"
         max-file: "3"
-    # If you want your db to persist in dev
+    # Database data - use 'docker compose down -v' to wipe
     volumes:
-     - "./data/postgres:/var/lib/postgresql/data"
+      - postgres_data:/var/lib/postgresql/data
 
   database-migration:
     build:
@@ -194,3 +195,4 @@ volumes:
   hardhat_deployments:
   ignition_deployments:
   hardhat_snapshots:
+  postgres_data:
--- a/.env.example
+++ b/.env.example
@@ -104,7 +104,7 @@ VITE_FINALITY_WINDOW_APPEAL_FAILED_REDUCTION = 0.2 # 20% reduction per appeal fa
 VITE_MAX_ROTATIONS = 3
 
 # Set the compose profile to 'hardhat' to use the hardhat network
-COMPOSE_PROFILES = 'hardhat'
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -132,6 +132,7 @@ services:
         max-file: "3"
 
   postgres:
+    profiles: ["local"]
```
cristiam86: @kstroobants I think this should be configured the other way around. Make it retrocompatible so current users don't need to add the "local" profile to run the Studio in the same way as before

kstroobants: How? I don't see a different way.

The local profile is part of the compose profiles in the environment file so the local development user can still run docker-compose up and it will use the postgres db.

cristiam86: I meant that maybe the profile should be something like "remote-db"
1. just `local` doesn't says about what service, "db" points to the database.
2. As far as I understand, after this is merged, developers running the studio locally will need to add this profile to their docker compose up command. Istead, if we state the profile "remote-db" to activate the remote database, developers won't need to do anything, we will need to set this profile in the infra pipelines.

kstroobants: 1. i can call it local-db
2. yes local users need to add this profile to the docker compose up command or add it to the compose profiles in the env file. In the infra pipeline we do not want to use this db because we use a gcp managed one thats why this service is targeting the local users. The remote database does not live in the docker-compose.yml file.

cristiam86: if you call it local-db we are in the same problem, developers will need to change the way they run the studio. I don't want this, it is a breaking change, needs to be updated in the docs, communicated, etc.

Instead, I want the profile to be "remote-db" meaning that, when it is set, the studio runs pointing to the GCP database instead of the local database. Then we will need to add this to the infra repo to be sure that the hosted studio in the GCP virtual machines runs with the remote-db enabled

kstroobants: I understand what you want but I do not see a way how to program it with docker. Docker does not have a negation statement. This does not work:
```
postgres:
  profiles: ["!remote-db"] 
```
 
Current solution:
- Local devs have COMPOSE_PROFILES=local in their env, so docker-compose up works. It is part of the env.example file so why is it a breaking change?
- GCP/prod doesn't get postgres service started.

(Note: the same was done in the explorer)

PR #1290, opened 2025-09-02
feat(rpc): add gen_getContractCode endpoint; store accounts data a…

epsjunior:
### What
- Add RPC endpoint `gen_getContractCode` to return contract code from `current_state.data.state.accepted[BASE64(get_code_slot())]`.
- Register the endpoint in `register_all_rpc_endpoints`.
- Update `AccountsManager.create_new_account_with_address` to initialize `data` as `{}` (JSONB object) instead of `"{}"` (string), avoiding str/dict mismatches.

### Why
- Align code retrieval with the canonical storage layout used by GenVM.
- Eliminate errors when `data` is a string (e.g., "'str' object has no attribute 'get'").
- Provide a simple, reliable way for clients to fetch contract code for tooling (schema generation, debugging, etc.).

### Testing done
- Called `gen_getContractCode` for a deployed contract; received Base64-encoded code and decoded to bytes with `b64ToArray`.
- Verified returned code works with `gen_getContractSchemaForCode` after converting to 0x-hex.
- Verified undeployed/empty state returns JSON-RPC error “Contract not deployed.”.
- Created a new account and confirmed `current_state.data` is a JSONB object (`{}`), not `"{}"`.

### Decisions made
- Server strips the 4-byte little-endian length prefix stored in the code slot and returns only raw code bytes (Base64).
- No backfill migration included; legacy rows with `"{}"` remain. Optional DB update can be handled in a separate PR if needed.

### Checks
- [x] I have tested this code

### Reviewing tips
- Focus on `backend/protocol_rpc/endpoints.py` (new `get_contract_code` and registration) and `backend/database_handler/accounts_manager.py` (data initialization).
- Confirm `get_code_slot` is used and the value is looked up under `data.state.accepted[BASE64(slot)]`.
- Check that only Base64 code bytes are returned (no length prefix).

### User facing release notes
- New JSON-RPC method `gen_getContractCode(address)` returns the contract’s code as Base64.
- New accounts now store `data` as a JSON object, improving compatibility and preventing errors in downstream reads.

## Summary by CodeRabbit

* **New Features**
  * Added an RPC endpoint to retrieve a contract’s deployed code by address, returning the code as base64 for wallets, explorers, and dev tools.

* **Bug Fixes**
  * Default account state for new addresses now initializes as an empty object to avoid inconsistencies.
  * Normalized legacy contract state formats to reliably detect and extract deployed code.

### DIFF
--- a/backend/database_handler/accounts_manager.py
+++ b/backend/database_handler/accounts_manager.py
@@ -42,7 +42,7 @@ class AccountsManager:
             return existing_account
 
         # If account doesn't exist, create it
-        account = CurrentState(id=address, data="{}", balance=0)
+        account = CurrentState(id=address, data={}, balance=0)
         self.session.add(account)
         self.session.commit()
         return account
--- a/backend/database_handler/contract_snapshot.py
+++ b/backend/database_handler/contract_snapshot.py
@@ -1,7 +1,8 @@
 # database_handler/contract_snapshot.py
 from .models import CurrentState
 from sqlalchemy.orm import Session
-from typing import Optional
+from typing import Optional, Dict
+import base64
 
 
 class ContractSnapshot:
@@ -14,7 +15,7 @@ class ContractSnapshot:
     contract_address: str
     contract_code: str
     balance: int
-    states: dict[str, dict[str, str]]
+    states: Dict[str, Dict[str, str]]
 
     def __init__(self, contract_address: str | None, session: Session):
         if contract_address is not None:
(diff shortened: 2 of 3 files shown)


backend/protocol_rpc/endpoints.py
```diff
@@ -428,6 +428,59 @@ async def get_contract_schema_for_code(
     return json.loads(schema)
 
 
+def get_contract_code(accounts_manager: AccountsManager, contract_address: str) -> str:
+    if not accounts_manager.is_valid_address(contract_address):
+        raise InvalidAddressError(
+            contract_address,
+            "Incorrect address format. Please provide a valid address.",
+        )
+    contract_account = accounts_manager.get_account_or_fail(contract_address)
```
kstroobants, line 437: Can you use backend/database_handler/contract_snapshot.py
Make a class object with the contract_address. Then make a function in that class to obtain the code from accepted state with that code slot and call that function on the object in this rpc method?

backend/database_handler/accounts_manager.py
```diff
@@ -42,7 +42,7 @@ def create_new_account_with_address(self, address: str) -> Account:
             return existing_account
 
         # If account doesn't exist, create it
-        account = CurrentState(id=address, data="{}", balance=0)
+        account = CurrentState(id=address, data={}, balance=0)
```
kstroobants, line 45: did you run the studio and execute a deploy transaction to see if it breaks something?

epsjunior: yes, its working properly

backend/protocol_rpc/endpoints.py
epsjunior: done

backend/protocol_rpc/endpoints.py
```diff
@@ -428,6 +428,39 @@ async def get_contract_schema_for_code(
...
+            data = json.loads(data)
+        except Exception:
+            raise InvalidAddressError(contract_address, "Contract not deployed.")
+    if not data:
+        raise InvalidAddressError(contract_address, "Contract not deployed.")
+
+    state = (data.get("state") or {}) if isinstance(data, dict) else {}
+    code_b64 = ContractSnapshot.extract_deployed_code_b64_from_state(state)
+    if not code_b64:
+        raise InvalidAddressError(
```
AgustinRamiroDiaz, line 457: nit: probably this is not an `InfalidAddressError` and actually something like `NotFound`

backend/database_handler/contract_snapshot.py
```diff
@@ -65,3 +66,45 @@ def _load_contract_account(self, session: Session) -> CurrentState:
...
+            if not stored:
+                return None
+
+            raw = base64.b64decode(stored, validate=True)
+            if len(raw) < 4:
+                return None
+
+            code_len = int.from_bytes(raw[0:4], byteorder="little", signed=False)
+            if len(raw) < 4 + code_len:
+                return None
```
AgustinRamiroDiaz, line 106: It'd be nice to have the explanation stored alongside the code so we don't forget
```suggestion
            # strips the 4-byte little-endian length prefix stored in the code slot and returns only raw code bytes (Base64)
            raw = base64.b64decode(stored, validate=True)
            if len(raw) < 4:
                return None

            code_len = int.from_bytes(raw[0:4], byteorder="little", signed=False)
            if len(raw) < 4 + code_len:
                return None
```

PR #1294, opened 2025-09-02
fix: resolve PostgreSQL constraint violation for duplicate transaction hashes

mpaya5:
# What

- **Simplified `get_transaction_count()` method** to always use PostgreSQL database as source of truth instead of inconsistent Hardhat RPC calls
- **Implemented robust retry logic** in `backend/protocol_rpc/endpoints.py` with proper rollback handling and fresh nonce recalculation for hash collision scenarios
- **Added `get_transaction_hash()` method** in `backend/protocol_rpc/transactions_parser.py` to extract original transaction hashes from Metamask-signed transactions when possible
- **Cleaned up codebase** by removing unused database pool configuration, session managers, and associated test files
- **Simplified database engine configuration** to use basic pool settings instead of complex managed database optimizations

  # Why

  

  To fix the critical PostgreSQL error `duplicate key value violates unique constraint "transactions_pkey"` that was blocking transaction processing when users attempted to resend transactions from Metamask. The root cause was:

  1. **Incorrect nonce usage** in hash generation causing identical hashes for different transactions
  2. **Nonce inconsistencies** between Hardhat's internal nonce tracking and PostgreSQL transaction records
  3. **Lack of collision handling** when duplicate hashes were detected
  4. **Over-engineered pool configuration** that added complexity without solving the core issue

  This PR ensures reliable transaction processing by using PostgreSQL as the single source of truth for nonce calculation and implementing proper collision recovery.

  # Testing done

  

  - ✅ **Manual testing**: Verified that resending transactions from Metamask no longer causes PostgreSQL constraint violations
  - ✅ **Hash collision testing**: Confirmed that when duplicate hashes are detected, the system properly rolls back and recalculates with fresh nonce
  - ✅ **Database consistency**: Verified that `get_transaction_count()` correctly returns counts from PostgreSQL transactions table
  - ✅ **Edge case handling**: Tested both `duplicate key value violates unique constraint` and `PendingRollbackError` scenarios

  # Decisions made

  

  - **PostgreSQL as single source of truth**: Decided to always use database transaction count instead of Hardhat RPC for nonce calculation to eliminate inconsistencies
  - **Fresh nonce recalculation**: After database rollback, recalculate nonce from current database state rather than using static increments
  - **Code simplification**: Removed complex database pool configurations and session managers that weren't addressing the core hash collision issue
  - **Preserve Metamask hashes when possible**: Use original transaction hashes from Metamask, only regenerate when collision detected

  # Checks

  - [x] I have tested this code

  # Reviewing tips

  

  Key areas to focus on:
  - **Line 243 in `transactions_processor.py`**: The critical fix changing `current_nonce` to `nonce` in hash generation
  - **Lines 811-891 in `endpoints.py`**: The new collision detection and retry logic with rollback handling
  - **Simplified `get_transaction_count()`**: Now consistently uses PostgreSQL instead of Hardhat RPC
  - **Removed files**: Check that elimination of complex pool configs and session managers doesn't break other functionality

  # User facing release notes

  

  **Fixed Critical Transaction Processing Bug**

  - ✅ **Resolved "duplicate key constraint" errors** when resending transactions from Metamask
  - ✅ **Improved transaction reliability** with better nonce handling and collision recovery  
  - ✅ **Simplified system architecture** by removing unnecessary complexity in database connection handling
  - ✅ **Enhanced error handling** for edge cases during transaction submission

  Users should no longer experience PostgreSQL constraint violation errors when attempting to resend or retry transactions from their Metamask wallet.

## Summary by CodeRabbit

- Refactor
  - Transaction nonce is now derived from the database, removing reliance on blockchain RPC and associated connectivity/error paths.
  - Standardized nonce progression across environments by making the database the source of truth.

- Tests
  - Updated unit tests to validate database-backed transaction counting.
  - Added scenarios for checksum success/failure and various count outcomes.
  - Removed RPC-dependent tests and assertions.

### DIFF
--- a/backend/database_handler/accounts_manager.py
+++ b/backend/database_handler/accounts_manager.py
@@ -42,7 +42,7 @@ class AccountsManager:
             return existing_account
 
         # If account doesn't exist, create it
-        account = CurrentState(id=address, data="{}", balance=0)
+        account = CurrentState(id=address, data={}, balance=0)
         self.session.add(account)
         self.session.commit()
         return account
--- a/backend/database_handler/contract_snapshot.py
+++ b/backend/database_handler/contract_snapshot.py
@@ -1,7 +1,8 @@
 # database_handler/contract_snapshot.py
 from .models import CurrentState
 from sqlalchemy.orm import Session
-from typing import Optional
+from typing import Optional, Dict
+import base64
 
 
 class ContractSnapshot:
@@ -14,7 +15,7 @@ class ContractSnapshot:
     contract_address: str
     contract_code: str
     balance: int
-    states: dict[str, dict[str, str]]
+    states: Dict[str, Dict[str, str]]
 
     def __init__(self, contract_address: str | None, session: Session):
         if contract_address is not None:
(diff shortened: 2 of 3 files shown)


cristiam86 (changes requested):
- review names
- test only the nonce part
- if we still have the error, move the transaction retry logic to the transactions_processor

PR #1441, opened 2026-02-12
feat: forward LLM error details to transaction receipts

cristiam86:
# What

- Enhanced `llm.lua` to capture structured error info (HTTP status, model, provider, error message) from LLM provider failures and pass it through `lib.rs.user_error()` ctx
- Added `parse_ctx_from_module_error_string()` parser in `error_codes.py` to extract `primary_error` / `fallback_error` from the Rust debug format string that GenVM produces for `INTERNAL_ERROR`
- Added `ctx` field to `GenVMInternalError` exception so error context flows through the Python exception chain
- Added `_extract_llm_error_result()` to `Receipt` to surface LLM errors as `llm_error_result` in the transaction receipt dict
- Fixed `raw_error` construction in `node/base.py` to include `ctx` and use correct `"fatal"` key

# Why

- When an LLM provider fails (invalid API key, rate limiting, etc.), the error details were lost during the GenVM → Python error propagation. The `result_data` arrives as a Rust debug format string (not a Python dict), so the structured ctx containing error details was silently discarded.
- This made it impossible for users and operators to understand _why_ a validator failed — the receipt only showed a generic "NO_PROVIDER_FOR_PROMPT" error code without the HTTP status, provider name, or error message.

# Testing done

- [x] Added 7 unit tests for `parse_ctx_from_module_error_string` covering: field extraction, nested Maps, escaped quotes, Nil values, integer status conversion
- [x] All 295 backend unit tests pass
- [x] All 31 DB/SQLAlchemy integration tests pass
- [x] All 103 frontend unit tests pass

# Decisions made

- Used character-by-character brace matching (not regex) for extracting nested `Map({...})` content from Rust debug strings, since nested structures break regex-based approaches
- Only extract `status`, `model`, `provider`, `error_message` fields — these are the actionable fields users need. The `body` field (which may contain large nested error payloads) is deliberately excluded from `llm_error_result`
- The `llm_error_result` field is `None` when no LLM errors occur, keeping receipts clean for successful transactions

# Checks

- [x] I have tested this code

# Reviewing tips

- Start with `backend/node/llm.lua` to see how error info is now captured at the source
- Then follow the propagation: `error_codes.py` (parser) → `genvm/base.py` (provide_result) → `node/base.py` (receipt construction) → `types.py` (receipt serialization)
- The Rust debug format parsing in `error_codes.py` is the core new logic — the test cases show the format clearly

# User facing release notes

- Transaction receipts now include detailed LLM error information (`llm_error_result`) when a validator's LLM provider fails, showing the HTTP status code, provider name, model, and error message

## Release Notes

* **New Features**
  * Enhanced error reporting with detailed context information, including status codes and error details for improved diagnostics.

* **Improvements**
  * Improved system reliability through increased health check retry attempts.
  * Optimized transaction processing performance with eager data loading.
  * Enhanced error handling and logging for better visibility into system failures.

### DIFF
--- a/backend/protocol_rpc/health.py
+++ b/backend/protocol_rpc/health.py
@@ -296,35 +296,31 @@ async def _run_health_checks() -> None:
         _health_cache.genvm_healthy = genvm_ok
         _health_cache.genvm_error = genvm_error
 
-        # Best-effort parse of permit info (do not fail overall health on missing keys).
+        # Best-effort parse of permit info from /status.
+        # The manager returns: {"permits": {"current": N, "max": M}, "executions": {...}}
+        # where permits.current = available permits (semaphore value).
         try:
-            max_permits = genvm_status.get("max_permits")
-            current_permits = genvm_status.get("current_permits")
-            active_exec = genvm_status.get("active_executions")
+            permits_obj = genvm_status.get("permits") or {}
+            executions_obj = genvm_status.get("executions")
 
-            max_permits_i = int(max_permits) if max_permits is not None else None
-            current_permits_i = (
-                int(current_permits) if current_permits is not None else None
-            )
+            max_permits_i = int(permits_obj["max"]) if "max" in permits_obj else None
+            # permits.current IS the available count (semaphore value)
             available_i = (
-                max_permits_i - current_permits_i
-                if max_permits_i is not None and current_permits_i is not None
-                else None
+                int(permits_obj["current"]) if "current" in permits_obj else None
             )
+            active_i = len(executions_obj) if isinstance(executions_obj, dict) else None
 
             _health_cache.genvm_max_permits = max_permits_i
-            _health_cache.genvm_current_permits = current_permits_i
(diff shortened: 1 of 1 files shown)


MuncleUscles:
I'm not sure those errors belong in the receipt, I think logs is the right place
- Receipt - for developers/users of the network
- Logs - for manager of the studio

cristiam86:
> I'm not sure those errors belong in the receipt, I think logs is the right place
> 
> - Receipt - for developers/users of the network
> 
> - Logs - for manager of the studio

You are right, the intention here is to be able to access them fast in the studio explorer

cristiam86:
> I'm not sure those errors belong in the receipt, I think logs is the right place
> 
> * Receipt - for developers/users of the network
> * Logs - for manager of the studio

Removed from the first level of the receipt and also fixed a race condition in the triggered transactions integration tests due to the query optimization that you did recently

PR #1539, opened 2026-03-17
fix: add partial indexes for worker claim queries and explorer lookups

cristiam86:
## Summary
- Adds 7 targeted partial indexes on the `transactions` table to eliminate sequential scans identified in CloudSQL slow query reports
- Partial indexes for the 3 worker claim queries (`claim_next_finalization`, `claim_next_appeal`, `claim_next_transaction`) — only include active/in-flight rows so they stay small
- Index for the `NOT EXISTS` anti-join subquery used in all 3 claim queries (`to_address, blocked_at WHERE blocked_at IS NOT NULL`)
- Explorer indexes: `to_address + created_at DESC` for contract transaction lookups, `from_address` for `COUNT(DISTINCT)` in health checks, `created_at DESC` for time-range queries
- Uses regular `CREATE INDEX` (not `CONCURRENTLY`) because the migration must commit prior enum values before creating partial indexes that reference them, and `CONCURRENTLY` cannot run inside a transaction

## Test plan
- [ ] Run DB/SQLAlchemy tests: `docker compose -f tests/db-sqlalchemy/docker-compose.yml --project-directory . run --build --rm tests`
- [ ] Apply migration on staging and verify indexes exist with `\di transactions*` in psql
- [ ] Confirm CloudSQL slow query log no longer shows seq scans for the claim queries
- [ ] Verify explorer page load times improve (query 4: full SELECT by to_address)

## Summary by CodeRabbit

* **Performance**
  * Added database indexes to speed up claim-related and explorer queries.
  * Faster listing and filtering of transactions (pending, appealed, address-based and recent activity), leading to quicker page loads and more responsive UI.
  * Improved backend query efficiency for transaction views and dashboards, reducing latency for users.

### DIFF
--- a/explorer/src/app/api/health/route.ts
+++ b/explorer/src/app/api/health/route.ts
@@ -1,0 +1,5 @@
+import { NextResponse } from 'next/server';
+
+export async function GET() {
+  return NextResponse.json({ status: 'ok' });
+}


cristiam86:
Re: CodeRabbit's comment about `CREATE INDEX CONCURRENTLY` vs regular `CREATE INDEX`:

Good catch. We originally used `CREATE INDEX CONCURRENTLY` with `autocommit_block()`, but it failed in CI for two reasons:

1. The partial indexes reference enum values (`LEADER_TIMEOUT`, `VALIDATORS_TIMEOUT`) added by earlier migrations. When running all migrations from scratch, PostgreSQL requires those enum values to be committed first (`UnsafeNewEnumValueUsage`).
2. We need an explicit `COMMIT` before index creation to flush those enum values, which means we're managing the transaction boundary manually — `CONCURRENTLY` requires full autocommit mode which conflicts with this.

Since this is a dev sandbox (Studio) with small tables, blocking `CREATE INDEX` during migration is acceptable. Updated the PR description to reflect the actual behavior.

PR #1674, opened 2026-06-10
feat(ci): bind GenVM to a branch — source-build path + main host-protocol adaptation

MuncleUscles:
## What

Adds a second way to bind Studio's images to GenVM — building from a git ref instead of a release tarball — and adapts Studio's host protocol to GenVM `main`, so CI can track GenVM tip ahead of releases.

### Two GenVM sources (mutually exclusive build args)

| Arg | Use | Path |
|---|---|---|
| `GENVM_TAG` | Releases / pinned CI (default, unchanged) | Downloads `genvm-linux-<arch>.tar.xz` + `genvm-runners-all.tar.xz` from the GitHub release, post-install downloads the executor |
| `GENVM_REF` | Dev binding to a branch/SHA | New `nixos/nix` builder stage clones `genlayerlabs/genvm` at the ref and replicates genvm's own release recipe (`.github/workflows/release.yaml` + `support/ci/pipelines/build.sh`): patch `flake-config.json` with the resolved SHA, `runner-script.py dependencies prefetch-needed`, then `nix build .#manager-<arch>-linux`, `.#executor-<arch>-linux`, `.#runners-all`, merged into the same `/genvm` layout the tarballs produce. Post-install runs with `--default-download false` (nothing to download for an unreleased SHA). |

Prebuilt runner tars (fixed-output derivations) are preloaded from genvm's public GCS cache (`runner-script.py download --nix-preload --allow-partial`, same trick as genvm's `support/ci/pipelines/test-rust.sh`), so `runners-all` normally copies instead of compiling. The builder layer cleans sources + nix store afterwards so the cached layer is roughly the size of `/genvm`.

Exactly one of `GENVM_TAG` / `GENVM_REF` must be set — enforced in the Dockerfiles with a clear error; `docker-compose.yml` now passes both through (previously hard-required `GENVM_TAG`).

### Binding model: codebase default (`third_party/genvm/version`) < env override (e2e matrix / release train)

**No repo variables are involved.** The GenVM binding is a checked-in file, `third_party/genvm/version` (same convention as genlayer-node's), versioned together with the protocol-compatibility code — on `v0.123-dev` it says `main`; at release cut it gets flipped to the release tag in the same history that pins the protocol.

Resolution in `load-test-oha.yml` / `backend_integration_tests_pr.yml` ("Resolve GenVM binding" step):
1. If `GENVM_TAG` or `GENVM_REF` is already set in the environment (injected by the e2e pipeline / release train), it wins. Both set = hard error.
2. Otherwise read `third_party/genvm/version`: value matching `^v[0-9]` → `GENVM_TAG` (release tarball path); anything else → `GENVM_REF` (source-build path). Missing/empty file = hard error, no guessing.
3. Any resulting `GENVM_REF` (override or file) is resolved to a commit SHA via `gh api repos/genlayerlabs/genvm/commits/<ref>`.

**Cache strategy:** in ref mode the buildx GHA cache moves into `scope=genvm-<sha>` (and the genvm-precompile `actions/cache` key becomes `ref-<sha>`), so image layers and precompile caches invalidate exactly when genvm moves — the expensive source build happens once per genvm commit, cached for every run after that. Tag mode keeps today's cache behavior.

Because the file on this branch says `main`, **this PR's CI validates itself** — the integration/load runs build genvm from source with no out-of-band setup.

### Protocol adaptation (rc3 → main)

Against rc4+/main, the genvm manager rejects Studio's `/genvm/run` payload with
`500 unknown variant 'execution_budget_per_round', expected one of: Internal, External` (e.g. run 27292725606) — GenVM #310 (CON-549, `ef04f7f4`) plus the rc4 restructure changed the fee-allocation schema.

**Request — `message_fee_allocation`** (genvm refs: `executor/crates/common/src/domain/fees.rs`, `modules/implementation/src/manager/run.rs`, mirror in `tests/runner/origin/fees.py`):

```text
OLD (rc3, flat parent-pointer list)              NEW (main, nested tree)
{                                                {
  message_type: "InternalAccepted"                 recipient, call_key, budget,
    | "InternalFinalized" | "External",            on: "accepted" | "finalized",
  parent_index: int | None,                        fee_params:
  recipient, call_key, budget,                       {"Internal": {leader_timeunits_allocation,
  fee_params: {                                        validator_timeunits_allocation,
    leader_timeunits_allocation,                       execution_budget_per_round, rotations,
    validator_timeunits_allocation,                    max_price_gen_per_time_unit,        # new (CON-549)
    execution_budget_per_round,                        storage_fee_max_gas_price,          # new
    rotations,                                         receipt_fee_max_gas_price}}         # new
  }                                                  | {"External": {gas_limit, max_gas_price}},
}                                                  children: [...]   # replaces parent_index
                                                 }
```

Non-root allocations are now sent nested under their roots (previously dropped), so genvm can ABI-encode the correct `subtree` for each matched emission. The three price caps are sourced from Studio's `fees_distribution` — they are not part of Studio's canonical per-node 5-field feeParams ABI, which is unchanged.

**Response — emissions** (`executor/crates/common/src/domain.rs`): `PostMessage`/`DeployContract`/`EthSend` now carry `fee_params` as a struct and an ABI-encoded allocation `subtree` (8-field internal params ABI per `domain/fees/abi.rs`). `backend/node/genvm/base.py` translates these back at the boundary: internal params re-encoded to Studio's canonical 5-field ABI bytes, external to the 2-field ABI, `subtree` decoded into the canonical `allocationSubtree` node list (inner 8-field feeParams converted to 5-field so subtree-equality checks hold). `fees.consumed` and all canonical RPC fee surfaces are unchanged. Absent fields (older genvm, stubbed tests) keep today's defaults.

`backend/node/genvm/origin/` is re-synced with genvm main's `tests/runner/origin/` (new vendored `fees.py`).

### Drive-by fixes (pre-existing, found by the test suite)

Three `TestMessageFeeModes` tests (from the original v0.6 fee accounting, #1649) were failing on `v0.123-dev` — regressions from `b88e0594` ("enforce studio fee execution floor"):
- `create_child_fee_accounting` zeroed Mode-1 children's message-fee bucket (no allocation subtree ⇒ refund) instead of seeding the declared-budget remainder as the child's bucket.
- `fill_message_fee_payload_from_allocation` phase-checked (`onAcceptance`) external allocations; external messages have no accepted/finalized lifecycle (genvm's `matches_external` ignores `on` for the same reason).

## Validation

- `tests/unit` full suite: **894 passed** (incl. rewritten genvm-payload tests and a new emission round-trip test exercising main's dict `fee_params` + ABI `subtree`).
- `tests/consensus/test_payable_scenarios.py`: all passing, including the 3 previously-failing `TestMessageFeeModes` tests. (Rest of `tests/consensus` fails identically before/after — pre-existing fixture issues, unrelated.)
- **Local end-to-end on Apple Silicon (native linux/arm64, no emulation):** full backend image built with `GENVM_REF=main` — genvm stage 7m20s (`GENVM_BUILD_CORES=6`, 16 GB Docker VM, no OOM); executor reports `genvm v0.3.0-devef04f7f-aarch64-linux-release`. Compose stack (jsonrpc + 3 consensus-workers) healthy; **Counter deploy FINALIZED and 2/2 concurrent increments preserved** (the `test_state_integrity` flow); a fee-bearing `gltest` Storage run against the stack produced the first measured `--fee-profile` (see PR comments).
- amd64 source build: validated in this PR's own CI (the binding file says `main`, so the integration/load workflows exercise the full source-build path per run).

## Rollout — read before merging

- [ ] After this PR, Studio speaks GenVM main's protocol; rc3-era tarballs reject the new fee-allocation payload the same way rc4 rejected the old one (deploys stall in PROPOSING). No release tag bridges the gap — genvm's calldata decoder rejects unknown struct fields (`executor/crates/calldata/tests/derive_decode_excess_fields.rs`), so the 7-field `Internal` params only decode at/after genvm #310 (`ef04f7f4`, main, unreleased). The binding file says `main` accordingly; don't pin a tag older than the CON-549 schema.
- [ ] First CI run per genvm SHA pays the full source build (Rust manager+executor; runners come prebuilt from GCS). Watch the runner's disk/time on the first build; subsequent runs hit the `scope=genvm-<sha>` layer cache.
- [ ] At release cut (and once a GenVM release ≥ the CON-549 schema ships), flip `third_party/genvm/version` to the release tag — the same file flip the release train already does for genlayer-node. The e2e matrix / release train can override per-run by exporting `GENVM_TAG` / `GENVM_REF`.


MuncleUscles:
### Local validation results (Goal A — source-build stage, `GENVM_REF=main`)

Ran on Apple Silicon (Docker Desktop, 16 GB VM) — findings:

**What passed**
- Builder stage end-to-end logic: clone @ `main`, `flake-config.json` patch (SHA + `v0.0.0-dev<short>` executor version), `runner-script.py dependencies prefetch-needed` (incl. mirror fallback when `zlib.net` 404'd), flake evaluation, and dependency derivation builds.
- **GCS runner preload + `runners-all`** (the riskiest piece — whether unreleased main runner hashes exist as prebuilt artifacts): all 19 runner FODs for current `main` (`ef04f7f4`) preloaded from the public `gh-af` bucket via `runner-script.py download --nix-preload`, after which `nix build .#runners-all` completed **with zero compilation** and produced the expected `runners/<id>/<xx>/<hash>.tar` layout. Validated natively on aarch64-linux, which only works because nothing had to compile — confirming the preload short-circuit works as designed.

**What's deferred to CI (native amd64)**
- The manager/executor Rust compile. Two upstream constraints make it impossible to finish on this machine:
  1. genvm's pinned rust toolchain ships x86_64 binaries only (`.cargo-wrapped: cannot execute` on aarch64-linux) — matches genvm's README "production build supported only on x86_64 linux hosts". The Dockerfile documents amd64 as the supported platform.
  2. Under Rosetta emulation the build OOM-killed the nix daemon in a 16 GB VM even with `cores = 4` (took BuildKit down with it). Hence the new `GENVM_BUILD_CORES` build arg, plus `filter-syscalls = false` (nix's seccomp filter can't load under emulation).
- The compile itself is exactly what genvm's release CI runs per release (`nix build .#manager-amd64-linux` / `.#executor-amd64-linux` on ubuntu-latest), so risk is low — but the first `vars.GENVM_REF` CI run should be watched for runner disk/time.

**Protocol adaptation (Goal C)**
- `tests/unit`: 894 passed (incl. new emission round-trip test against main's dict `fee_params` + ABI `subtree`).
- `tests/consensus/test_payable_scenarios.py`: all green, incl. the 3 `TestMessageFeeModes` cases that were failing on `v0.123-dev` (regressions from `b88e0594`, fixed here).
- Compose stack + Counter deploy (`test_state_integrity` flow) against a main-built genvm: left for CI — emulated wasmtime locally isn't representative. Suggest setting `vars.GENVM_REF=main` on this PR (repo variables apply to PR runs) and re-running the Load Test / backend integration workflows to validate the full pipeline pre-merge.

MuncleUscles:
### Local end-to-end validation, round 2 (binding-file model, native arm64)

Design change since the first comment: the repo-variable mechanism is gone — the binding is the checked-in `third_party/genvm/version` (`main` on this branch), env `GENVM_TAG`/`GENVM_REF` override it (e2e matrix / release train). This PR's CI now exercises the source-build path on its own.

**Native arm64 build (Apple Silicon, no emulation)** — the earlier "arm64 impossible" conclusion was wrong; two real bugs were hiding behind it, both fixed here:
1. genvm upstream: `support/rust.nix` patchelf's its rust toolchain with a hardcoded **x86-64** ELF loader → `.cargo-wrapped: cannot execute` on aarch64-linux (genvm CI never sees it; arm64 release targets are cross-compiled from x86_64 runners). The builder stage rewrites it to `ld-linux-aarch64.so.1` on arm64, guarded to skip once fixed upstream. *Worth a genvm issue/PR: pick the interpreter from the host platform instead of hardcoding.*
2. post-install's lief RPATH rewrite SIGABRTs on the zig/musl-linked **amd64** executor (`Can't read got entries` / `Node not found. Can't resize the section`). Source-built outputs already carry `$ORIGIN` rpaths from genvm's own nix `patch-rpath`, so the source path now runs post-install with `--bin-patch false` (bin-check still validates the binaries execute).

**Results on this machine (M-series, 16 GB Docker VM):**
- Full backend image with `GENVM_REF=main`: genvm builder stage **7m20s** (`GENVM_BUILD_CORES=6`, no OOM; manager ~2.5m, executor ~2.5m, runners preloaded from GCS, nix GC keeps the layer small). Executor: `genvm v0.3.0-devef04f7f-aarch64-linux-release`.
- Compose stack: jsonrpc + 3 consensus-workers healthy.
- **Counter deploy (state-integrity flow): deploy FINALIZED, 2/2 concurrent increments preserved, `get_count() == 2`** — the rc3→main host-protocol adaptation works live (no `unknown variant 'execution_budget_per_round'` 500s).
- **First measured fee profile** (genlayer-testing-suite #94 `--fee-profile`, fee-bearing Storage deploy + `update_storage`, receipts read at FINALIZED, default headroom 1.25 over observed `executionConsumed` 314536/314514):

```json
{
  "version": 1,
  "network": "localnet",
  "measuredAt": "2026-06-10T23:09:51Z",
  "deploy": {
    "executionBudgetPerRound": "393170",
    "totalMessageFees": "0"
  },
  "methods": {
    "update_storage": {
      "executionBudgetPerRound": "393143",
      "totalMessageFees": "0"
    }
  }
}
```

Notes for the fee-profile workflow: receipts only expose non-zero `fees.consumed` once **FINALIZED** (profiling with the default ACCEPTED wait records zeros), and only fee-bearing transactions carry a `fees` object at all — gasless runs produce an empty profile (the reporter warns accordingly). The #94 branch also needs the fee-capable genlayer-py (`fees=`/`wait_until` API, v0.19-dev line) — its PyPI floor `genlayer-py>=0.18` is too low.

MuncleUscles:
### Integration-test failure analysis (run 27311560490) + fix (`07d39253`)

The backend integration `test` job failed while the Load Test pipeline (same images, same commit) passed — every failure was an `@mock_validators` case stalling in PROPOSING. From the compose log dump:

```
genvm manager /genvm/run failed: 500 {'error': "modules are required but not running (is_sync=false with 'n' permission)"}
```

**This is a race, not a protocol regression.** Studio stops+restarts the genvm llm/web modules on *every* validator-change event (1130 `Stopped LLM module` lines in one run — the mock-validator tests reconfigure 5 validators right before each deploy). A `/genvm/run` landing between stop and start gets that 500; studio treated it as fatal, and with the churn the consensus-level retries kept colliding → tx stuck in PROPOSING until the 420 s test timeout. The manager-side check is **identical in v0.3.0-rc3** (`modules/implementation/src/manager/handlers.rs:85`) — genvm main just has a wider module-restart window, turning a rare flake into a near-deterministic one. Nondet deploys in the same run succeeded whenever they missed the window (e.g. LlmErc20 at 23:44), confirming timing.

**Fix:** `backend/node/genvm/origin/base_host.py` now routes only this transient 500 (and the sibling "not all are running") into the existing `GENVM_MANAGER_RUN_RETRIES` retry loop (default 3 attempts, exp backoff, env-tunable) via a dedicated `GenVMManagerRetryableError`; all other non-200s keep failing fast — a protocol-shaped 500 like the old `unknown variant` must never be retried. Retry stats record `retry_reason: manager_modules_not_running`. Two parametrized regression tests added (transient 500 → retried to success; other 500 → immediate fail). `tests/unit`: **897 passed**.

Follow-up worth considering (out of scope here): debounce studio's module stop/start churn — restarting the llm module once per validator *snapshot* (or only when the provider config actually changed) instead of per event would shrink this window for every genvm version.

MuncleUscles:
## Integration suite: root causes + fixes (runs 27311560490 / 27314907925)

The ~37 failures had **three independent causes** — none flaky, all reproduced and fixed:

### 1. Host-protocol deadlock → every nondet tx stuck in `PROPOSING` (the timeout class)

GenVM main widened fuel values to U256 (32-byte LE) in the host wire protocol; our `base_host.py` still used rc3's 8-byte framing in two places:

- `REMAINING_FUEL_AS_GEN` reply: we sent 8 bytes, the executor `read_exact`s 32 → it blocked on the missing 24 bytes while our host loop blocked waiting for the next method — a silent read-read deadlock. It sits directly in the `exec_prompt`/nondet path (`wasi/genlayer_sdk.rs` fetches remaining fuel right before calling the LLM module), which is why **only nondet executions** (wizard_of_coin, llm_erc20, football_prediction_market, genvm_smoke) hung until `LEADER_TIMEOUT` (10 min) while every deterministic test passed. Confirmed against the deployed executor commit (`v0.3.0-devef04f7f`, `executor/src/host/mod.rs::remaining_fuel_as_gen`).
- `CONSUME_FUEL` request: executor writes 32 bytes, we read 8 → stream desync whenever invoked.

This also explains why the retry fix (07d39253) didn't move the needle: the hang produces no 500 — the manager accepted the run and the executor was alive, just deadlocked.

Locally on this branch's stack: wizard_of_coin went from a 10-minute hang to **passing in ~11s** after the 2-line width fix.

### 2. Floating runner aliases now resolve to the genvm-main runner (the `NameError: gl` class)

`py-genlayer:test` / `py-genlayer:latest` / `py-genlayer-multi:test` resolve (under `--debug-mode`) to the newest runner — now the genvm-main one (`1zr6nqk5…`, bootloader-based), whose SDK **no longer star-exports `gl`** (`import genlayer as gl` + `gl.contract.Contract` is its idiom). Every fixture still on an alias died at import with `NameError: name 'gl' is not defined` in the leader receipt; the upgrade-contract suite then cascaded into `Contract not found`.

Fix: pinned all affected fixtures to the v0.3.0-rc3 runner every green contract already uses (`py-genlayer:1jb45aa8…`, `py-genlayer-multi:06zyvrl…`): 8 icontracts fixtures, the 5 inline upgrade-test fixtures, and `multi_file_contract/runner.json`. Contract bodies were already valid v0.3 SDK — no behavioral changes, no assertion changes.

### 3. `log_indexer` pinned a vanished embeddings runner

`py-lib-genlayer-embeddings:09h0i209…` no longer exists in the image (executor → `contract_error: invalid_contract`). Repinned to `0bmbm3cy…` from the current registry.

### Also included

- `TEST_JSONRPC_URL` override for `tests/common/request.py` / conftest / upgrade tests (the previous hardcoded `localhost:4000` made the suite untestable against a non-default stack; CI behavior unchanged — the variable is unset there).

### 3b. Worker fleet dies on `NO_PROVIDER_FOR_PROMPT` (found while verifying)

The llm module reports `NO_PROVIDER_FOR_PROMPT` as fatal; Studio's fatal-leader circuit breaker then **stops the claiming worker**. Mock-validator tests reconfigure validators before every test (module stop/start storm), so one affected tx serially killed all 3 workers → queue stalled silently (`worker_task_died`, restart_count 0). Fixed: that cause keeps the worker alive (tx still has its bounded 3-cycle retry budget → CANCELED) plus a short backoff (`GENVM_TRANSIENT_FATAL_BACKOFF_S`, default 3s) before the reset.

### Local verification (stack built from this branch, mock validators)

| Test | Before | After |
|---|---|---|
| wizard_of_coin | 10-min PROPOSING hang | **pass, ~11s** |
| company_naming, football, llm_erc20, oracle_factory, multi_read/multi_tenant/read_erc20, multi_file, utf8_roundtrip | NameError / hang | **pass** |
| test_upgrade_contract (31 tests) | all error (NameError cascade) | **pass** |
| genvm_smoke | hang | not verifiable locally — its empty mock_response means a real LLM call, and this stack has no `OPENROUTERAPIKEY`; CI has one (it passed there pre-genvm-main) |
| log_indexer | invalid_contract | contract + schema verified via direct RPC (slow locally: embeddings model load in schema/exec path) — CI is the arbiter |

29/31 of the previously-failing set green locally; the two holdouts are local-env limitations, not code issues.

### Notes / leftovers

- `error_execution_contract.py`, `error_llm_contract.py`, `error_web_contract.py` still use `py-genlayer:test` — out of scope here (excluded from CI via `-m "not error_handling"`), but they'll need the same repin when that marker is re-enabled.
- Studio-side observation while debugging: a tx whose pinned validator set was wiped mid-flight produces fatal `LLM_NO_PROVIDER`, which stops the claiming worker by design; with all workers stopped the queue stalls silently ("worker_task_died", restart_count 0). Worth a follow-up issue — recovery exists (3-cycle escalation to CANCELED) but costs all worker tasks.

MuncleUscles:
## Direction change: contracts upgraded FORWARD to the genvm-main SDK

Per Edgars: repinning fixtures to the rc3 runner (098bbb8a) was the wrong direction — the point of binding studio to genvm main is that the test contracts move to the latest SDK. The previous CI run (27323792167) is still useful evidence: **894 passed / 1 failed**, confirming the two backend fixes (U256 fuel framing dfdee12c, worker circuit-breaker 7c72f32b) hold in CI; the single failure was log_indexer on the rc3/old-embeddings pairing.

### What changed now

**Every integration fixture contract is upgraded to the genvm-main SDK idiom** (refs: genvm `runners/genlayer-py-std/src/genlayer/`, `tests/cases/semi-stable/llm/woc.py`, `tests/cases/stable/runners/multi-file/`):

- `import genlayer as gl` + `from genlayer import *` (star-import provides types only; `gl` is no longer star-exported)
- `class X(gl.contract.Contract)` (was `gl.Contract`)
- `gl.contract.get_at` / `gl.contract.deploy` (was `gl.get_contract_at` / `gl.deploy_contract`)
- `raise gl.vm.UserError(...)` (was `gl.Rollback`, which no longer exists)
- `gl.eq_principle.prompt_comparative(fn, principle)` — `principle` is positional-only now
- `@allow` from `genlayer.storage` (was `@allow_storage`)
- embeddings: `VecDB[T, S, V, Distance]` (4th type param), `EuclideanDistanceSquared`
- `gl.nondet.exec_prompt(p, response_format='text')` mirroring genvm's own contracts

**Pins are concrete main hashes** (never `test`/`latest`):
- `py-genlayer:1zr6nqk597d97kg0dyxg0shhrykx5v02zjgnyrajapy4wlqvfvwh`
- `py-genlayer-multi:0jw6688wgnbqghr2wwg2mqkf2ryb5if7w6bp5zcwbq6zlks5ybg4`
- `py-lib-genlayer-embeddings:0wcvi35grdr47ynkckzriz5sjn5080w2njk7v2cqx3xpp6p1989y`

**Backward-compat canary:** `examples/contracts/user_storage.py` stays deliberately on the rc3 runner (`py-genlayer:1jb45aa8…`, old SDK idiom) with a prominent comment — `test_user_storage` now doubles as a regression test that old-runner contracts keep executing on the genvm-main executor. (Gotcha found while setting it up: the executor parses the entire leading comment block as version+runner JSON, so the canary note must sit after a blank line.)

**gltest discovery shim** (`tests/integration/gltest_compat.py`, activated in conftest): gltest 0.24.0 AST-discovers contracts only by the legacy `gl.Contract` base; upstream main already matches `gl.contract.Contract` but it's unreleased. The shim patches the two discovery functions to accept both. Remove when the `genlayer-test` pin moves past 0.29.2.

**One more backend host-adaptation gap fixed** (found via log_indexer): GenVM main reports storage changes for readonly (view) executions — e.g. VecDB lazy-init inside a view `knn` — and `provide_result` applied them into a readonly state proxy, tripping `assert not self.readonly` → `-32603 Internal error` on `gen_call`. View-execution storage changes are now discarded.

### Local verification (stack from this branch, mock validators)

47 of 48 relevant tests green locally:
- full `test_upgrade_contract` suite (28) with main-SDK inline fixtures
- all icontracts: company_naming, football_prediction_market, intelligent_oracle_factory, llm_erc20, log_indexer, multi_file, multi_read_erc20, multi_tenant_storage, read_erc20, utf8_roundtrip, wizard_of_coin, storage, faucet (2), payable_escrow (4)
- canary `test_user_storage` (old runner on main executor)
- genvm_smoke is the one exception — not verifiable on this machine (its empty mock_response means a real LLM call; no `OPENROUTERAPIKEY` locally — CI has one and genvm_smoke passed there on the previous run)

Two genvm-main quirks found and worked around (worth flagging to the genvm team):
1. cover-tree `knn` from a **write** method trips the deterministic-mode float trap (`wasm_trap DeterministicMode`); the same call from a view works. log_indexer's `update_log` now locates entries by iteration instead.
2. similarity values changed (the main embeddings runner ships `EuclideanDistanceSquared`; the retired `09h0…` runner used a cosine-style metric) — test windows recalibrated to the new deterministic values.

PR #1676, opened 2026-06-11
test(ci): gasless-mode integration coverage

MuncleUscles:
## Why

Rally's hosted studio runs in **gasless mode** — all three price envs set to zero:

- `GENLAYER_STUDIO_GEN_PER_TIME_UNIT=0`
- `GENLAYER_STUDIO_STORAGE_UNIT_PRICE=0`
- `GENLAYER_STUDIO_RECEIPT_GAS_PRICE=0`

(`StudioFeePolicy.fee_accounting_enabled()` derives from them.) Until now only unit tests covered this configuration; no integration/CI job ever ran a zero-price stack, so a fee-path change could silently break the hosted deployment.

## What

This pins the gasless contract end to end:

- **`tests/integration/test_gasless_mode.py`** (marker `@pytest.mark.gasless`, excluded from the default sweep via `pytest.ini` `addopts`, mirroring the `error_handling` convention):
  - `sim_getFeeConfig` reports `enabled: false` and all-zero policy prices (hard-fails if the stack under test is not actually gasless),
  - deploy + fee-less `update_storage` write succeed (the pinned SDK has no fee kwargs — exactly how gasless clients submit),
  - canonical receipt `fees` is `null` (via `eth_getTransactionByHash`),
  - reads work,
  - `sim_estimateTransactionFees` (the RPC genlayer-js `estimateTransactionFees` hits) succeeds and does not mutate state,
  - tolerant-behavior pin: passing explicit fee params (`fees.distribution` + `feeValue`) to the estimate surface on a gasless stack is **accepted** (validated against a zero-price policy with required deposit 0), not rejected.

- **`.github/workflows/backend_integration_tests_pr.yml`**: new `test-gasless` job — same bake/compose pipeline as `test` (same GenVM resolve step and buildx GHA cache scope, cache **reads only** so it never races the main job's cache writes), but with the three price envs forced to `0` in `.env` before `compose up`, a readiness step that asserts `sim_getFeeConfig.enabled == false`, and a serial `gltest ... -m gasless` run of only the gasless module. The main `test` job is untouched.

## Verification

- Ran the module locally against a zero-price compose stack (arm64, GENVM_REF source build): both tests pass.
- Verified the default `tests/integration/` sweep deselects the gasless tests and `-m gasless` selects exactly the 2 new ones.

### DIFF
--- a/.github/scripts/install-python-dependencies.sh
+++ b/.github/scripts/install-python-dependencies.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+python -m pip install --only-binary :all: -r requirements.txt
+python -m pip install --only-binary :all: -r requirements.test.txt
+python -m pip install --only-binary :all: -r backend/requirements.txt
--- a/.github/workflows/backend_integration_tests_pr.yml
+++ b/.github/workflows/backend_integration_tests_pr.yml
@@ -35,6 +35,7 @@ jobs:
     env:
       PYTHONPATH: ${{ github.workspace }}
       COMPOSE_PROFILES: hardhat
+      GENVM_CACHE_DIR: /tmp/genvm-cache
 
     steps:
       - name: Checkout code
@@ -90,8 +91,6 @@ jobs:
           echo "GENVM_REF=$GENVM_REF_VALUE" >> "$GITHUB_ENV"
           grep "^GENVM_TAG=" .env
           grep "^GENVM_REF=" .env
-
-      # TODO: we should also add also heuristai and anthropic keys to the e2e tests and test all providers
 
       - name: Configure LLM provider
         env:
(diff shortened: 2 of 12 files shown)


MuncleUscles:
**SonarCloud status note**

The quality gate flags the new `test-gasless` job. The 3 security *hotspots* from the first analysis (tag-pinned docker actions, cleartext curl) were fixed in fb6b8cf — hotspot count is now 0.

The remaining failure is `new_security_rating = C` from 7 issues on lines that are byte-for-byte copies of the existing `test` job (Sonar counts them as new code because the job is new):

- `pip install -r requirements*.txt` without `--only-binary`/hash locking (S8541/S8544, lines 350-352) — same as the main job's install step; hash-locking the three requirements files is out of scope for this PR.
- `chmod -R 0777 /tmp/genvm-cache` (S2612, line 384) — intentional: the GenVM precompile cache must be writable by the container user's UID, identical to the main job.

These can be marked *Accepted* in the SonarCloud UI by a project admin; functionally the new job introduces no security surface the existing pipeline doesn't already have.

**CI verdict**: `test-gasless` green twice (21.5 min cold GenVM cache, 5.5 min warm; the gasless pytest run itself is ~16 s), main `test` suite green both rounds (65 passed, gasless tests correctly excluded from the default sweep), Load Tests green on re-run (first failure was a runner out-of-disk during the cold GenVM nix source build, unrelated to this diff).

PR #1697, opened 2026-07-01
Port studio to genvm-manager (v0.3)

kp2pml30:
Depends-On: https://github.com/genlayerlabs/genlayer-studio/pull/1697
Depends-On: https://github.com/genlayerlabs/genlayer-py/pull/96
Depends-On: https://github.com/genlayerlabs/genlayer-js/pull/192
Depends-On: https://github.com/genlayerlabs/genlayer-testing-suite/pull/98

Ports studio to genvm-manager (v0.3).

NOTE: I did not touch linters, gltest and other stuff

## Compatibility with old stuff

1. I checked that deployed studio is v0.2.16, so in new manager I added version for v0.2.16 with all runners, old layout and other stuff
2. I changed callldata abi for method call: `"method"` key became `""` so that it's value is prefix of binary calldata
  a. v0.2.16 has a hard remaping for exposing to contracts "method" and giving back only `""`
  b. currently manager if sees `"method"` in input calldata will remap it to `""` automatically and print an error
  c. **migration of genlayer-js and genlayer-py is needed!** migration in node and studio is more or less handled, I believe
3. code slot changed, this poses issues and I have not supported full v0.3 <-> v0.2 interop, I may need your help
  a. you may be interested in https://github.com/genlayerlabs/genvm-manager/blob/02741163aa89b10331f1585d9ea42868a816945f/docs/website/src/impl-spec/appendix/manager-api.yaml#L256
  b. maybe for real db we can add a field for each contract alongisde storage or whatever

## Other notes

there will be hash breaking commit to v0.3 with changing runners for allocating from balance

I have not yet supported builds releasesd and other stuff, as we are not merged into core and not even near that. For that reason I uploaded a hardcoded x86_64 artifact for you builded form commit for convenience

I ran **some** tests, deployed UI and verified basic functionality working. However, more complicated stuff may not work, it is a draft of changes

### DIFF
--- a/examples/contracts/log_indexer.py
+++ b/examples/contracts/log_indexer.py
@@ -1,22 +1,22 @@
-# v0.3-dev
+# v0.3.0
 # {
 #   "Seq": [
-#     { "Depends": "py-lib-genlayer-embeddings:1md4i1njqn0h0psgjdl97mz10rpp1268ychpn6l2dmr81fbvxknb" },
-#     { "Depends": "py-genlayer:1zr6nqk597d97kg0dyxg0shhrykx5v02zjgnyrajapy4wlqvfvwh" }
+#     { "Depends": "py-lib-genlayer-embeddings:kr2rb2dcp01mw9khpg3tg2jasx4f82mcsy3eg08rjj1zdcm9q350" },
+#     { "Depends": "py-genlayer:9b8kjyda2ycxyq4ea6g4yfpnydxhd52gqba5rb8dw7krkh5mn9p0" }
 #   ]
 # }
 
 import numpy as np
 import genlayer as gl
-from genlayer import *
-from genlayer.storage import allow
+from genlayer.types import *
+from genlayer.storage import TreeMap
 import genlayer_embeddings as gle
 
 from dataclasses import dataclass
 import typing
 
 
-@allow
+@gl.storage.allow
 @dataclass
 class StoreValue:
     log_id: u256
@@ -52,7 +52,7 @@ class LogIndexer(gl.contract.Contract):
                 continue
             if log_id not in self.log_vector_ids:
                 continue
-            if self.log_vector_ids[log_id] != u32(result.id):
+            if self.log_vector_ids[log_id] != result.id:
                 continue
             return {
                 "vector": list(str(x) for x in result.key),
@@ -64,7 +64,7 @@ class LogIndexer(gl.contract.Contract):
 
     @gl.public.write
     def add_log(self, log: str, log_id: int) -> None:
-        key = u256(log_id)
+        key = log_id
         if key in self.log_vector_ids:
             self.vector_store.get_by_id(self.log_vector_ids[key]).value = StoreValue(
                 text=log, log_id=key
@@ -73,11 +73,11 @@ class LogIndexer(gl.contract.Contract):
 
         emb = self.get_embedding(log)
         vector_id = self.vector_store.insert(emb, StoreValue(text=log, log_id=key))
-        self.log_vector_ids[key] = u32(vector_id)
+        self.log_vector_ids[key] = vector_id
 
     @gl.public.write
     def update_log(self, log_id: int, log: str) -> None:
-        key = u256(log_id)
+        key = log_id
         if key in self.log_vector_ids:
             self.vector_store.get_by_id(self.log_vector_ids[key]).value = StoreValue(
                 text=log, log_id=key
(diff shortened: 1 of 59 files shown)


examples/contracts/log_indexer.py
```diff
@@ -1,22 +1,21 @@
...
-from genlayer.storage import allow
+from genlayer.types import *
 import genlayer_embeddings as gle
 
 from dataclasses import dataclass
 import typing
 
 
-@allow
+@gl.allow
```
kp2pml30: should be `@gl.storage.allow`

MuncleUscles:
**Re: code slot / v0.2↔v0.3 interop** — read your notes, here's the decision.

First: yes, deployed contracts on hosted studio (v0.2.16) have to survive the upgrade. That's the constraint everything below follows from.

Your per-contract metadata suggestion is the right call, and I'd make it explicit rather than inferred from storage. Sniffing the layout can't work: `sha3(zeros ++ u32(1))` is `CODE` in the v0.2 layout but `CONTRACT` in v0.3 (`_RootOffsets`), so that address is populated in *both* layouts with different meanings — presence tells you nothing. Explicit version it is.

Concretely:

1. **`genvm_major` column on `current_state`** (contract rows), alembic migration backfilling all existing rows to `2` — safe, since everything deployed today is v0.2.16 by definition. Same shape as the old `balance` column migration; notably we have zero precedent for rewriting the `data` JSONB itself, and we shouldn't start now.
2. **Stamp new deploys via the manager's `POST /contract/detect-version`** (the manager-api.yaml section you linked) — one canonical version-detection implementation instead of studio growing its own.
3. **Version-keyed layout registry in one module**: `root_offsets_for(major)` / `get_code_slot(major)`, i.e. `{2: RootOffsets(CODE=1, …), 3: RootOffsets(MAJOR=0, CONTRACT=1, CODE=2, …)}` — no scattered `if major == 2` branches.
4. **Thread the contract's major through the six touch points**: deploy write (`node/base.py::_put_code_to`), consensus (`consensus/base.py` → `decisions.py`), the upgrade tx type=3 path (`worker.py:~1491` — writes the code slot directly, don't miss it), `contract_snapshot.py::extract_deployed_code_b64`, execution-time `storage_read`, schema extraction.
5. **No slot migration.** v0.2 contracts stay on the v0.2 layout forever and run on the legacy executor you added — rewriting their code into the new slot would break exactly the contracts we're preserving.
6. **The major travels with snapshots/forks/appeal rollback**, otherwise a forked v0.2 contract gets read with v0.3 offsets.
7. **Upgrade tx (type=3): cross-major upgrades are rejected.** The upgrade path validates the new bytecode via `/contract/detect-version` against the stored `genvm_major` and rejects on mismatch with an explicit error — storage is incompatible across majors on both the root layout (same slot addresses, different meanings) and the SDK's own data encoding, so a cross-major upgrade could only silently corrupt. v0.2 contracts remain upgradable with v0.2 code; moving to v0.3 = redeploy. (Separate question: what's the `UPGRADERS` root slot in v0.3 for — is there an in-band migration story planned for future majors?)

Since v0.3 storage self-describes via the `MAJOR` slot, this whole thing is a one-time compatibility layer for the pre-MAJOR era — the column is the permanent record for legacy contracts, and everything new is versioned in-band.

Also 👍 on deriving the slot from `root_offsets` — the old hardcoded `b"\x01\x00\x00\x00"` next to an unused `CODE_SLOT_OFFSET` constant was a trap.

kp2pml30:
Note on `test_llm_invalid_unknown_key` (in `tests/integration/icontracts/tests/test_error_llm.py`):

This test and the rest of `test_error_llm.py` are **identical to the base branch `v0.123-dev`** — this PR only adds a one-line clarifying comment. It is marked `@pytest.mark.error_handling`, which `tests/integration/pytest.ini` deselects by default (`-m "not error_handling and not gasless"`), and **no CI job passes `-m error_handling` on either branch**, so it is not exercised by CI here or on base. i.e. the situation is the same on the base branch; nothing about this test is introduced by the port.

When run manually with `-m error_handling`, it requires `UNKNOWN_KEY` to be **defined** in the env (to any non-empty invalid value). If the var is undefined, the v0.3 templater resolves `${ENV[UNKNOWN_KEY]}` to `""` → provider disabled → zero providers → deploy CANCELED (`num_of_rounds=0`, no receipt), so the assertion cannot find its receipt. With `UNKNOWN_KEY` set to a bogus value the provider initializes, the LLM call 401s, and the test passes (ERROR receipt + `GenVM internal error` in stderr + `Leader Timeout`) — verified locally.

kp2pml30:
genvm can't be built in docker. Building genvm runners requires sandbox, which requires privileges for docker

PR #1716, opened 2026-07-28
feat(genvm): calls between different majors of genvm

kp2pml30:
This PR adds support for

1. Pinning a version for a contract (commit 1)
  a. Messages between "majors" work out of the box with it
2. Calls between contracts (commit 2) via rewrite of the manager interface to support multiple host <-> executor connections
  a. Nondet and some other stuff is forbidden on it, there are sharp edges
  b. Referral contract should work
3. Docker build of GenVM (commit 3: sandbox, priveleges, etc; that slack message)

## How to Get Rally Work

NOT tested on real rally, follow up may be needed, but on migrating real DB set new param to `"re:^v0\.2\."` for existing contracts)

## Disclaimer

Vendored genvm communication got diverged from origin, I ported retries and other stuff back to genvm, but this needs agent re-review (I did many iterations already, but...)

Diff is much less scary:

```sh
❯ git diff (git merge-base HEAD v0.123-dev) --stat -- ':(exclude)*test_*.py' ':(exclude)backend/node/genvm/origin/*'
 .genvm-nix-closure/.gitkeep                                                                 |   0
 .github/actions/genvm-runners-closure/action.yml                                            |  65 ++++++++++++++++
 .github/workflows/backend_integration_tests_pr.yml                                          |  26 +++++++
 .github/workflows/load-test-oha.yml                                                         |  13 ++++
 .gitignore                                                                                  |   4 +
 backend/consensus/base.py                                                                   |  11 +++
 backend/consensus/decisions.py                                                              |   2 +
 backend/database_handler/contract_processor.py                                              |   2 +
 backend/database_handler/contract_snapshot.py                                               |   5 ++
 backend/database_handler/migration/versions/c1d2e3f4a5b6_add_reroute_to_to_current_state.py |  30 ++++++++
 backend/database_handler/models.py                                                          |   5 ++
 backend/database_handler/snapshot_manager.py                                                |   6 +-
 backend/domain/types.py                                                                     |   7 ++
 backend/node/base.py                                                                        |   7 ++
 backend/node/genvm/base.py                                                                  | 347 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---------------------
 backend/protocol_rpc/endpoints.py                                                           |  51 ++++++++++++-
 docker/Dockerfile.backend                                                                   |  12 +++
 docker/Dockerfile.consensus-worker                                                          |  12 +++
 docker/scripts/download_genvm.sh                                                            |  46 ++++++++----
 tests/db-sqlalchemy/contract_reroute_to_test.py                                             |  62 ++++++++++++++++
 third_party/genvm/version                                                                   |   2 +-
 21 files changed, 613 insertions(+), 102 deletions(-)
```

## Changeset

Depends-On: genlayerlabs/genlayer-node#1647
Depends-On: genlayerlabs/genlayer-studio#1716
Depends-On: genlayerlabs/genvm-manager#9

## Summary by CodeRabbit

* **New Features**
  * Added Studio-only executor version pinning for contract deployments.
  * Preserved executor routing settings across contract state, snapshots, and restarts.
  * Enabled cross-contract calls between contracts running on different executor versions.
  * Added validation to ensure routing pins use supported version formats and debug modes.

* **Bug Fixes**
  * Improved executor connection recovery, retries, cleanup, and completion handling.
  * Updated GenVM assets to version v0.6.0-rc1.

* **Tests**
  * Added comprehensive coverage for routing, persistence, nested execution, retries, and connection lifecycles.

### DIFF
--- a/backend/consensus/base.py
+++ b/backend/consensus/base.py
@@ -319,6 +319,15 @@ def node_factory(
     )
 
 
+def transaction_genvm_executor_selector(transaction: Transaction) -> str | None:
+    """Studio-only GenVM executor override carried by the transaction."""
+    return (
+        transaction.sim_config.genvm_executor_selector
+        if transaction.sim_config
+        else None
+    )
+
+
 def contract_snapshot_factory(
     contract_address: str,
     session: Session,
@@ -352,6 +361,9 @@ def contract_snapshot_factory(
         ret.contract_code = transaction.data["contract_code"]
         ret.balance = transaction.value or 0
         ret.states = {"accepted": {}, "finalized": {}}
+        # The contract row is still empty at deploy time, so the executor
+        # override can only come from the deploy transaction itself.
+        ret.genvm_executor_selector = transaction_genvm_executor_selector(transaction)
         return ret
 
     # Return a ContractSnapshot instance for an existing contract
@@ -2831,6 +2843,11 @@ class AcceptedState(TransactionState):
             ),
             to_address=context.transaction.to_address,
             leader_node_config=leader_receipt.node_config,
+            genvm_executor_selector=(
+                transaction_genvm_executor_selector(context.transaction)
+                if is_deploy
+                else None
+            ),
         )
 
         # Execute pre-effects (includes contract registration/update via executor)
--- a/backend/consensus/decisions.py
+++ b/backend/consensus/decisions.py
@@ -321,6 +321,7 @@ def decide_accepted(
     code_slot_b64: str | None,
     to_address: str,
     leader_node_config: dict,
+    genvm_executor_selector: str | None = None,
 ) -> tuple[list[Effect], list[Effect], ConsensusRound, ConsensusRound | None]:
     """Decide effects for AcceptedState.
 
@@ -419,6 +420,7 @@ def decide_accepted(
                             },
                         },
                     },
+                    "genvm_executor_selector": genvm_executor_selector,
                 }
                 pre_effects.append(RegisterContractEffect(contract_data=new_contract))
                 pre_effects.append(
--- a/backend/database_handler/contract_processor.py
+++ b/backend/database_handler/contract_processor.py
@@ -19,6 +19,10 @@ class ContractProcessor:
             self.session.query(CurrentState).filter_by(id=contract["id"]).one()
         )
         current_contract.data = contract["data"]
+        if "genvm_executor_selector" in contract:
+            current_contract.genvm_executor_selector = contract[
+                "genvm_executor_selector"
+            ]
         self.session.commit()
 
     def update_contract_state(
--- a/backend/database_handler/contract_snapshot.py
+++ b/backend/database_handler/contract_snapshot.py
@@ -17,6 +17,9 @@ class ContractSnapshot:
     contract_address: str
     balance: int
     states: Dict[str, Dict[str, str]]
+    # Executor version or `re:` selector this contract is pinned to, forwarded
+    # to the GenVM manager as `reroute_to`.
+    genvm_executor_selector: Optional[str] = None
 
     def __init__(self, contract_address: str | None, session: Session):
         if contract_address is not None:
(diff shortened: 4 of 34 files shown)


kp2pml30:
@coderabbitai review

MuncleUscles (changes requested):
The version-routing approach makes sense, but I think we need to address the upgrade path and a few manager lifecycle issues before merging.

### Upgrade compatibility

1. **Backfill existing v0.2 contracts in the migration.** Real Studios will be upgraded with many existing contracts, so adding a nullable column is not sufficient. For the supported upgrade path, existing deployed contract rows should receive a legacy selector such as `re:^v0\.2\.`, while preserving any non-null value and excluding EOAs. Please add a migration integration test that seeds a pre-migration v0.2 contract and verifies cross-version calls after upgrading.

2. **Automatically pin every successful new deployment.** At the moment the selector is persisted only when it was explicitly supplied. That leaves ordinary new contracts unpinned, which is why the v0.2 -> current case remains `xfail`. The manager's terminal result already reports the selected version line; please persist a selector derived from that (for example `re:^v0\.3\.`), then make the cross-version test pass in both directions.

3. **Use the same selector grammar as the manager.** Studio currently validates exact-looking versions, while the manager supports exact versions and `re:` selectors. Both submit-time validation and nested-call resolution need to accept and validate the same safe grammar, otherwise the migration backfill will fail at one of those boundaries.

### Manager lifecycle correctness

4. **Do not re-execute after receiving a terminal result.** Artifact retrieval or result decoding can fail after `Finished`, and that exception currently reaches the outer retry path, which can start a new execution with a new ID. Once a terminal event has been received, recovery should continue against that same run or fail without rerunning it. Please add a test where artifact retrieval or decoding fails after `Finished` and assert that no second run starts.

5. **Guard run operations across manager restarts.** The client tracks `(boot_id, genvm_id)`, but artifact retrieval, acknowledgement, and cancellation identify a run using only `genvm_id`. If the manager restarts and reuses an ID, stale cleanup could affect a new run. These operations should carry or verify the expected `boot_id`, or refuse to act after the manager generation changes.

6. **Handle empty or malformed consumed calldata explicitly.** `ConsumedResult.decode` handles `None`, but empty bytes or malformed payloads raise later. Please define the empty case explicitly and turn malformed data into a clear protocol error. In either case, it must not feed the execution replay path above.

### API and naming

7. **Rename the durable field.** `reroute_to` describes the manager's internal/debug override, but not what the persisted value means. Since this can contain either an exact version or a regex selector, `genvm_executor_selector` would be clearer for the database model, snapshots, and internal APIs. It can still be translated to `UnsafeOverrides(reroute_to=...)` at the manager boundary. The external `sim_config.reroute_to` name can remain temporarily for compatibility if needed.

8. **Reject this option on non-deployment transactions.** It is currently accepted for every transaction type but silently ignored outside deployment. Please reject it unless the transaction is a deployment, or move it into deployment-specific configuration.

### Non-blocking follow-ups

- Studio already supported GenVM source builds; the exported Nix closure is a CI acceleration/reproducibility mechanism. Please document that distinction. If local/CI unification is pursued, making the closure an optional cache input would avoid requiring developers to reproduce the host-Nix export workflow.
- Give the cancellation/drain logic in `close_connections` a timeout so shutdown cannot hang indefinitely.

kp2pml30:
> Automatically pin every successful new deployment. At the moment the selector is persisted only when it was explicitly supplied. That leaves ordinary new contracts unpinned, which is why the v0.2 -> current case remains xfail. The manager's terminal result already reports the selected version line; please persist a selector derived from that (for example re:^v0\.3\.), then make the cross-version test pass in both directions.

On the one hand I don't agree with it, because they should be auto-promoted, on the other I can already forsee a breaking change before clarke related to leader results, so I am not 100% sure. This flow is not normal and these pins should not exist on a "mainnet" / "clarke"

MuncleUscles (approved):
Approved. The migration backfill, selector handling, and manager lifecycle fixes look good. The earlier concern about pinning every new deployment does not apply: active development remains unpinned by design, with pinning performed at release time.

Issue #1721, opened 2026-08-10
Hardcoded Semaphore(8) in CommittingState serializes large validator committees


MuncleUscles:
## Problem

`CommittingState.handle` caps concurrent validator executions with a hardcoded `asyncio.Semaphore(8)` (`backend/consensus/base.py:1925`, introduced in #1272).

For large committees this dominates transaction latency: a 47-validator committee runs as ~6 sequential rounds of 8 (sliding window, but with similar per-validator LLM latencies it behaves like waves). This is the likely cause of the observed "47v is ~10× slower" behavior, and becomes increasingly relevant as appeals push committees to 23+.

## Why the cap exists (and why 8 is arbitrary)

Bounding GenVM concurrency is legitimate:
- each validator execution spawns a GenVM subprocess → unbounded concurrency risks OOM
- uvloop fd-registry collisions from concurrent GenVM socket operations (hit on the RPC path, fixed in #1648)
- DB connection-pool exhaustion

But the value **8** is not derived from anything — not container memory limits, host RAM, or committee size. It's a static default from #1272 with no recorded rationale. Meanwhile the RPC path's equivalent bound is already configurable via `GENVM_MAX_CONCURRENT` (`backend/protocol_rpc/endpoints.py:71`, default 8).

## Proposal

1. Make the consensus-path bound configurable — either reuse `GENVM_MAX_CONCURRENT` or add a dedicated env var (e.g. `CONSENSUS_VALIDATOR_CONCURRENCY`), defaulting to the current 8 for backward compatibility.
2. Deployments with adequate memory can then raise it to match committee sizes (23–47+) instead of paying a ~6× serialization penalty.

## References

- `backend/consensus/base.py:1925` — hardcoded `asyncio.Semaphore(8)`
- `backend/protocol_rpc/endpoints.py:67-73` — configurable RPC-path equivalent + rationale comment
- #1272 — introduced the semaphore (no rationale for the value)
- #1648 — documented fd/DB-pool failure modes of unbounded GenVM concurrency

PR #1723, opened 2026-08-10
feat(consensus): make validator execution concurrency configurable (v0.123)

MuncleUscles:
Refs #1721 — port of the same change targeting v0.121.

# What

- Replace the hardcoded `asyncio.Semaphore(8)` in `CommittingState.handle` with a module-level `VALIDATOR_MAX_CONCURRENT` read from the `CONSENSUS_VALIDATOR_MAX_CONCURRENT` env var (default 8, clamped to >= 1).
- Update the stale comment in `backend/protocol_rpc/endpoints.py` that referenced the hardcoded value.
- Document the new variable in `.env.example`.

# Why

A 47-validator committee currently runs through a fixed window of 8 concurrent validator executions (~6 sequential rounds), dominating transaction latency. The cap itself is legitimate (GenVM subprocess memory / fd / DB-pool bounds) but the value 8 was arbitrary and not tunable. Deployments with adequate memory can now size it to their committee sizes. Default behavior is unchanged.

# Testing done

- `python3 -m py_compile` on both edited modules.
- No behavior change with the variable unset (default 8).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -165,6 +165,9 @@ VALIDATORS_CONFIG_JSON='[
 JSONRPC_REPLICAS='1'
 CONSENSUS_WORKERS='3'
 MAX_PARALLEL_TXS_PER_WORKER='1'
+# Max validators executing concurrently per transaction (GenVM subprocesses).
+# Raise for large committees if memory allows.
+CONSENSUS_VALIDATOR_MAX_CONCURRENT='8'
 # Production Configuration (for Gunicorn deployment)
 WEB_CONCURRENCY='1'               # Number of Gunicorn workers (default: CPU cores * 2)
 # Service resources limit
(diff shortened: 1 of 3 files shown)


MuncleUscles:
Retargeted to `v0.123-dev` (rebased, same change). The prior E2E failures were not this PR: the v0.6 e2e train tests the fee-enabled stack (`sim_getFeeConfig` via genlayer-js v2 / cli v0.40 / wallet v0.2), which exists on `v0.123-dev` (fee accounting restored in 9e7bd54e) but not on the stale `v0.123` release branch — so all tooling components died on `Method not found: sim_getFeeConfig` regardless of this diff. Per the release flow, `v0.123` only receives promotions from `-dev` and shouldn't be a direct PR target.

/run-e2e studio

PR #1741, opened 2026-08-17
feat(rate-limit): meter cheap reads separately and expose usage headers

MuncleUscles:
# What

Three related changes to how `/api` traffic is rate limited.

**1. Cheap reads meter into their own bucket.** `RateLimitMiddleware` checked the quota before the JSON-RPC body was ever parsed, so every method cost the same. Reads that never enter the GenVM now meter into `ratelimit:{identity}:read:{window}` at `RATE_LIMIT_READ_MULTIPLIER` (default 10) times the tier limits. The standard bucket keeps its original key shape and numbers.

**2. `X-RateLimit-*` headers on every `/api` response**, plus on the 429 — `Limit`, `Remaining`, `Reset`, `Window`, `Bucket` — reported for whichever window is closest to exhaustion. Added to CORS `expose_headers`.

**3. `gen_getContractCode` no longer loads the whole contract state.** The code slot is extracted in SQL via `fetch_deployed_code_b64` instead of constructing a full `ContractSnapshot`.

New file `backend/protocol_rpc/rate_limit_methods.py` holds the read allowlist and body classification.

# Why

The limiter was method-agnostic, so `gen_getContractCode` — one indexed read — cost exactly what a write fanning out to LLM validators cost. That forces the limits to be sized for the expensive case, which starves the cheap read traffic that batch tooling generates. This surfaced as a user hitting 429s while fetching contract sources for automated reviews.

The headers exist because clients currently have no way to see how close they are to the ceiling — only `Retry-After` on rejection. The reporting user could not determine which window they had blown.

The query narrowing is coupled to change 1 on purpose: raising the read limit 10x without it would have traded an LLM-cost problem for a Postgres one, since `ContractSnapshot` pulls the entire `data` JSONB — every storage slot the contract owns — to read one deterministic slot out of it.

# Testing done

- 779 unit tests pass (24 new)
- `tests/unit/test_rate_limiter_lua.py` executes the Lua script against a stubbed `redis.call` implementing sorted-set semantics. The script previously had **no** real coverage — every test mocks `evalsha` — so a syntax error in it would have surfaced first in production on every `/api` request. Skips when no `lua` interpreter is present.
- `tests/unit/test_contract_code_fetch.py` covers nested/flat/legacy state shapes and compiles the generated SQL against the Postgres dialect, which mocks cannot validate.
- Classification is tested for GenVM-bound methods, mixed batches, and malformed/oversized bodies.

# Decisions made

- **The read allowlist is hand-maintained, not derived.** `DISABLE_INFO_LOGS_ENDPOINTS` looks like the natural source but contains `eth_call`, which routes through `_admit_genvm_call` and can fan out to LLM validators — reusing that list would have made the most expensive call in the system free. `gen_getContractSchema` and `gen_getContractSchemaForCode` are excluded for the same reason: both build a `Node` backed by a `GenVMManager`.
- **Ambiguity charges the stricter bucket.** Unparseable, oversized (>64KB), or a batch containing one expensive member all fall back to standard.
- **The standard bucket keeps its Redis key shape** so limits in flight at deploy time carry over rather than silently resetting.
- **`Reset` is time-until-oldest-entry-ages-out**, not the window length, which is the honest value for a sliding window.
- **Legacy and undeployed row shapes defer to `ContractSnapshot`** rather than reimplementing its error semantics in SQL.
- Reading the body in `BaseHTTPMiddleware` is safe here because Starlette 0.52.1 wraps the request in `_CachedRequest`, which replays the buffered body downstream. Verified in the installed source.

# Checks

- [x] I have tested this code

# Reviewing tips

- **The allowlist in `rate_limit_methods.py` is the highest-risk surface.** A method wrongly classified as cheap gets 10x capacity on a path that may cost real money. Worth checking each entry independently rather than trusting the grouping.
- This middleware fronts 100% of `/api` POST traffic, so the fail-open envelope in `dispatch` matters — body-parse failures must not reject requests.
- The DB win is partial and unmeasured: Postgres still detoasts the JSONB server-side, so what is saved is wire transfer and Python deserialization, not the read itself. If it benchmarks marginal, the 10x multiplier deserves a second look.
- Pre-existing and not addressed here: a batch array of N writes still counts as 1 request.

# User facing release notes

- API responses now include `X-RateLimit-Limit`, `X-RateLimit-Remaining`, `X-RateLimit-Reset`, `X-RateLimit-Window` and `X-RateLimit-Bucket`, so clients can pace themselves instead of discovering the limit by hitting it.
- Read-only RPC calls that do not execute contract code now have a substantially higher rate limit than contract execution calls.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -96,6 +96,7 @@ RATE_LIMIT_ENABLED='false'           # Enable/disable API key rate limiting
 RATE_LIMIT_ANON_PER_MINUTE='30'      # Anonymous (no API key) per-minute limit
 RATE_LIMIT_ANON_PER_HOUR='500'       # Anonymous per-hour limit
 RATE_LIMIT_ANON_PER_DAY='5000'       # Anonymous per-day limit
+RATE_LIMIT_READ_MULTIPLIER='10'      # Cheap reads (no GenVM) get this multiple of the tier limits
 
 # PENDING-tx queue depth caps for eth_sendRawTransaction (admission control).
 # Empty / unset = no cap (the default for self-hosted). Public shared
--- a/backend/database_handler/contract_snapshot.py
+++ b/backend/database_handler/contract_snapshot.py
@@ -1,6 +1,7 @@
 # database_handler/contract_snapshot.py
 from .models import CurrentState
 from .errors import ContractNotFoundError
+from sqlalchemy import func, select
 from sqlalchemy.orm import Session
 from typing import Optional, Dict
 import base64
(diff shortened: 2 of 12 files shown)


MuncleUscles:
## CodeQL `py/weak-sensitive-data-hashing` — dismissed as false positive

Alert [#187](https://github.com/genlayerlabs/genlayer-studio/security/code-scanning/187) fired on `hashlib.sha256(raw_key.encode())` in `backend/protocol_rpc/rate_limiter.py`, claiming SHA-256 is unsuitable because it is not computationally expensive.

**This PR does not touch that line.** `git blame` puts it at 93335ff7 (Feb 2026, #1445); it surfaced here only because changes to `check_rate_limit` altered the call path reaching it.

**The finding is also wrong on the merits.** The hashed value is not a password:

1. **It is a 256-bit CSPRNG token.** API keys are minted as `"glk_" + secrets.token_hex(32)` in `admin_create_api_key`. Slow KDFs (bcrypt, argon2, PBKDF2) exist to make brute-force and dictionary attacks expensive against *low-entropy, human-chosen* secrets. Against 2^256 of uniform randomness they buy nothing — the search is infeasible regardless of how fast the hash is.
2. **Fast hashing is a functional requirement.** This runs on the rate limiter's hot path, once per `/api` request. A deliberately expensive KDF would add roughly 100ms of latency to every call — degrading the exact thing this PR is trying to improve.
3. **Changing it is a breaking change.** Only the hash is persisted in `api_keys`; the raw key is unrecoverable by design. A different algorithm invalidates every key already issued.

SHA-256 over a high-entropy token is the standard construction for API key storage, and is what GitHub and Stripe do for the same reason.

Left open deliberately, as out of scope here: 8 × `py/stack-trace-exposure` (medium) look like they may be legitimate and are worth triaging separately.

PR #1755, opened 2026-09-03
fix: clear removed_log_ids tombstone when a log_id is re-added

ygd58:
## Summary

Fixes two related functional bugs in the `LogIndexer` example contract found while writing docs that reference this file ([genlayer-docs#449](https://github.com/genlayerlabs/genlayer-docs/pull/449)) — CodeRabbit's review of that docs PR flagged both, since the docs example is a byte-for-byte copy of this file.

## Bug 1: tombstone never clears on re-add

`remove_log` sets `removed_log_ids[key] = True` and deliberately leaves `key` in `log_vector_ids` (so the underlying `VecDB` entry can be reused rather than leaked on a later re-add).

But `add_log`/`update_log`'s "already indexed" branch (`if key in self.log_vector_ids`) never reset `removed_log_ids[key]` back to `False`. So a `log_id` that's removed and then re-added stayed permanently hidden from `get_closest_vector`'s tombstone filter, even though it was legitimately re-added.

## Bug 2 (deeper, found in follow-up review): stale embedding on update

The "already indexed" branch also only ever updated `VecDBElement.value` (the text/log_id metadata) in place. But `.key` — the actual stored *vector* that `knn()` ranks by — is a **read-only** property. So a log whose text changed via `add_log`/`update_log` stayed ranked by its **original** embedding forever, even though `.text` correctly reported the new content. This bug is independent of bug 1 — it affects any update to an existing `log_id`, tombstoned or not, and predates this PR.

## Fix

Both bugs share one fix: instead of patching `.value` in place, remove the stale `VecDB` element and insert a fresh one with a newly-computed embedding, updating `log_vector_ids` to the new id and clearing the `removed_log_ids` tombstone — unconditionally, in both `add_log` and `update_log`.

## Testing

Added two regression tests to `tests/integration/icontracts/tests/test_log_indexer.py`:
- Remove `log_id` 3, confirm it's no longer findable, re-add it via `add_log`, confirm it's visible again with its new text (bug 1).
- Update `log_id` 3 from one topic to a completely unrelated one (quantum computing vs. the original "third log" text) and confirm a query about the **new** topic correctly finds it — this is only possible if the vector was actually re-embedded, since the stale old embedding has nothing in common with the new topic (bug 2).

**Verified against a real running Studio instance** end-to-end (docker compose with `database-migration`, `jsonrpc`, `consensus-worker`, `TEST_WITH_MOCK_LLMS=true`) across three iterations as the fix was refined:

```
tests/integration/icontracts/tests/test_log_indexer.py::test_log_indexer PASSED
====================== 1 passed, 1 warning in 219.52s (0:03:39) =======================
```

### DIFF
--- a/examples/contracts/log_indexer.py
+++ b/examples/contracts/log_indexer.py
@@ -69,6 +69,7 @@ class LogIndexer(gl.contract.Contract):
             self.vector_store.get_by_id(self.log_vector_ids[key]).value = StoreValue(
                 text=log, log_id=key
             )
+            self.removed_log_ids[key] = False
             return
 
         emb = self.get_embedding(log)
@@ -82,6 +83,7 @@ class LogIndexer(gl.contract.Contract):
             self.vector_store.get_by_id(self.log_vector_ids[key]).value = StoreValue(
                 text=log, log_id=key
             )
+            self.removed_log_ids[key] = False
             return
 
         emb = self.get_embedding(log)
--- a/examples/contracts/log_indexer.py
+++ b/examples/contracts/log_indexer.py
@@ -66,29 +66,29 @@ class LogIndexer(gl.contract.Contract):
     def add_log(self, log: str, log_id: int) -> None:
         key = log_id
         if key in self.log_vector_ids:
-            self.vector_store.get_by_id(self.log_vector_ids[key]).value = StoreValue(
-                text=log, log_id=key
-            )
-            self.removed_log_ids[key] = False
-            return
+            # VecDBElement.key (the stored vector used by knn()) is
+            # read-only - only .value can be updated in place. Re-embedding
+            # requires removing the stale entry and inserting a fresh one,
+            # otherwise a changed log stays ranked by its old embedding.
+            self.vector_store.get_by_id(self.log_vector_ids[key]).remove()
 
         emb = self.get_embedding(log)
         vector_id = self.vector_store.insert(emb, StoreValue(text=log, log_id=key))
         self.log_vector_ids[key] = vector_id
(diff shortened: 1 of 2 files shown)


ygd58:
Ran this against a real Studio instance (docker compose + gltest, TEST_WITH_MOCK_LLMS=true) since I could not run integration tests in my sandbox. Confirmed passing:

```
tests/integration/icontracts/tests/test_log_indexer.py::test_log_indexer PASSED
1 passed, 1 warning in 142.34s
```

Updated the PR description accordingly.

ygd58:
Pushed a follow-up fix based on another CodeRabbit finding: the tombstone fix alone wasn't enough — `add_log`/`update_log`'s existing-ID branch only patched `.value`, never re-embedding, since `VecDBElement.key` (the stored vector) is read-only. A log whose text changed stayed ranked by its OLD embedding forever.

Fixed by removing the stale element and inserting a fresh one with the new embedding. Added a sharper regression test that updates a log to a completely unrelated topic and confirms search results follow the new content, not the old one (the earlier test used near-identical old/new text, which wouldn't have caught this).

I verified the *previous* commit (tombstone-only) against a real running Studio instance — passed. Have not yet re-verified this additional commit live, syntax-checked only. Will follow up once verified.

ygd58:
Confirmed passing end-to-end against a real Studio instance:

```
tests/integration/icontracts/tests/test_log_indexer.py::test_log_indexer PASSED
1 passed, 1 warning in 219.52s (0:03:39)
```

Both bugs (tombstone-never-clears and stale-embedding-on-update) are fixed and verified. Updated the PR description.

PR #1760, opened 2026-09-10
fix(rpc): commit funding credit before fund_account returns

kp2pml30:
## Problem

The `Load Tests` job hangs at `sim_fundAccount` until the 6h runner cap instead of finishing in ~13-17 min. The backend stops answering entirely, `/health` included.

## Mechanism

1. `fastapi_endpoint_generator.py:251` calls sync RPC handlers inline on the single uvicorn event loop — no threadpool
2. `fund_account` takes the `current_state` row lock for the funded address in `credit_tx_value_once`, then returns
3. The commit lives in `get_db_session`'s teardown, which runs *after* the response is awaited — the loop yields there
4. A concurrent request hits `INSERT … ON CONFLICT` on the same row and blocks inside psycopg2 (C code) → the event loop is frozen
5. The first request's teardown can never run → never commits → the row lock is held forever

Observed directly in `pg_stat_activity`: one backend `idle in transaction` for 10m29s on `UPDATE current_state SET balance = balance + …` with `blocked_by = {}`, and a second on `Lock/transactionid` blocked by it.

The bug is latent in the base branch. #1758 did not introduce it — before that commit a JSON-string amount raised `TypeError` in `credit_tx_value_once` before reaching `current_state`, so the request never took the lock.

## Fix

`session.commit()` at the end of `fund_account`, releasing the row lock before the handler returns.

**This is a release valve, not the structural fix.** Sync handlers still belong in a threadpool; that is a 21-handler change and does not belong in a hotfix.

## Verification

| Load | Before | After |
|---|---|---|
| `sim_fundAccount` n=100 c=10 (CI shape) | 600.6s, 100/100 timeout | **1.2s, 100/100 → 200**, p95 0.15s |
| `sim_fundAccount` n=4 c=2 (int amount) | 120.1s, 4/4 timeout | **0.1s, 4/4 → 200** |
| `idle in transaction` sessions after | 1, forever | 0 |

`pytest tests/unit` → 1569 passed, 7 skipped.

## Also here

`fix(ci): repin genlayer-py to compatible version` — v0.19.0-rc.2 is the first train build whose localnet preset uses chain ID 61127, matching `.env.example` and genlayer-js after #1756.

## Known, deliberately out of scope

1. Sync handlers on the event loop (the structural cause)
2. `lock_timeout` on the request-scoped session
3. Load harness: `Run Load Test - All Read Setup Endpoints` has no `timeout-minutes`, `load_test_all_read_setup_endpoints.sh:76` discards all oha output, and `check_test_result` passes on a single `[200]` rather than a full success count


MuncleUscles (approved):
Reviewed 13f0cdae1bfdae4004c1091e68fd4f864bae5756. The commit is correctly placed after transaction creation and balance credit, preserving their atomicity and completing the faucet transaction before the handler returns. No blocking issues found in this focused mitigation.

Load Tests, backend unit tests, and database integration tests are green. The main integration test is still running and E2E is queued, so this approval does not establish full merge readiness.

Non-blocking follow-ups: backport the funding fix to v0.121, add an explicit commit-before-return assertion to the funding regression coverage, and address synchronous dispatch separately.

Issue #1767, opened 2026-09-15
Consensus Solidity: unauthenticated internal-message emit helpers, ungated Transactions lifecycle, and ungated ConsensusManager seed writes


byfor8:
Hi team, reviewing the consensus contracts under hardhat/contracts/v2_contracts I found three missing-authorization issues in the same family. All are in this repository's source tree. I have not claimed them against Bradbury bytecode: the live ConsensusMain ABI does not contain emitTransactionAccepted, so I treat this tree as the studio/localnet deployment source. Static review only, no transactions were sent.

1) ConsensusMain._processInternalMessages is reachable from the external, unauthenticated emitTransactionAccepted / emitTransactionFinalized (ConsensusMain.sol:181-221). Each call stores arbitrary recipient addresses as ghost contracts via _storeGhost and enqueues transactions with attacker-chosen sender, recipient and data through _addTransaction. Anyone can poison the ghostContracts map and fabricate accepted/finalized events.

2) Transactions.activateTransaction, finalizeTransaction and submitAppeal are external without onlyGenConsensus, while sibling functions proposeTransactionReceipt, cancelTransaction and rotateLeader all carry the modifier (Transactions.sol:75-125, 331-351, 387-411). activateTransaction takes the activator address and the random seed as parameters and only compares them to storage, so a direct caller supplies both and effectively picks the committee seed. finalizeTransaction flips status to Finalized with no timeout check of its own, which bricks the later ConsensusMain.finalizeTransaction for that tx (it reverts TransactionNotAcceptedOrUndetermined). _calculateMinAppealBond hardcodes minAppealBond = 0 with a TODO (Transactions.sol:609-615), so appeals are free.

3) ConsensusManager.updateRandomSeedForRecipient / addNewRandomSeedForRecipient are external with no onlyGenConsensus, and initialize never sets the gate variable (ConsensusManager.sol:21-50). RandomnessUtils.updateRandomSeed only checks ecrecover(vrfProof) == _sender where _sender is itself a caller parameter (RandomnessUtils.sol:17-35), so anyone can sign the current seed with their own key and rotate any recipient seed. addNewRandomSeedForRecipient is first-writer-wins and ghost beacon addresses are CREATE-predictable, so future ghosts can be pre-seeded and permanently locked out of seed updates.

[...]

PR #1770, opened 2026-09-15
fix(consensus): ConsensusMain.addTransaction trusts caller-supplied _sender over msg.sender

### LINKED ISSUE #1768: ConsensusMain.addTransaction trusts caller-supplied _sender over msg.sender; Studio persists it as the Intelligent Contract caller
Hi team, one more from the same tree. _addTransaction only uses msg.sender when the caller passes _sender == address(0) (ConsensusMain.sol:241-260, 586-588):

    if (_sender == address(0)) {
        _sender = msg.sender;
    }

The Studio rollup path recovers the EVM signer of the raw transaction and checks it against decoded_rollup.from_address (backend/protocol_rpc/transactions_parser.py:514-542), but nothing requires that signer to equal the args.sender field that then becomes the Intelligent Contract caller: endpoints.py:2107-2238 stores from_address = args.sender via insert_transaction, and that value is what GenVM exposes as gl.message.sender_address. Fees are still debited from the real signer, so the attacker pays their own gas. The public ConsensusMain.addTransaction ABI takes _sender as its first argument, so the shape is exposed to any direct caller.

Impact: on Studio (and any deployment of this Solidity), anyone can submit a contract call that every Intelligent Contract sees as coming from any victim address. Every auth check that trusts gl.message.sender_address (token gates, owner-only contract methods, per-user balances) is bypassable. Verified on current main.

Fix: derive the sender exclusively from the recovered rollup signer / msg.sender, store that in insert_transaction, and reject _sender != msg.sender unless msg.sender is a registered ghost contract.

bars26:
Fixes #1768

# What

- `ConsensusMain.addTransaction` / `ConsensusMainWithFees.addTransaction`: reject `_sender != msg.sender` unless the caller is a registered ghost contract relaying its own caller (the legitimate `GhostBlueprint.addTransaction` relay pattern).
- `ConsensusMain._addFeeAwareTransaction` (the `fallback()`-routed v0.6 fee-aware entry point, added upstream after this branch was first opened): same check, applied to `senderWord`.
- `TransactionParser.get_genlayer_transaction`: use the verified rollup signer (`rollup_transaction.from_address`) instead of the unverified `args.sender` calldata field; raise `InvalidTransactionError` on mismatch.

# Why

`_sender` was accepted from the caller with no check against who actually signed the transaction. Anyone could sign a valid transaction from their own address while setting `_sender` to a victim's address, which then became the Intelligent Contract caller (`gl.message.sender_address`) — bypassing every auth check that trusts it (token gates, owner-only methods, per-user balances). The real signer only paid gas. While rebasing onto `v0.123-dev` I found the same unchecked pattern had been reintroduced by the new fee-aware fallback entrypoint, so that's fixed here too.

# Testing done

- Added 2 new unit tests to `tests/unit/test_transactions_parser.py` covering the matching-sender (legitimate) and spoofed-sender (rejected) cases.
- Full unit suite after rebasing onto `v0.123-dev`: `1580 passed, 7 skipped` (pre-existing skips, unrelated), `black --check` clean.
- Solidity: I could **not** verify with `npx hardhat compile`/hardhat tests — this repo's committed `hardhat/package.json` pins `hardhat-deploy@^2.0.0`, which requires `hardhat@^3.1.8` + `rocketh`, while `hardhat.config.js` and `deploy/001_deploy_fixture.js` are written for the old CommonJS/v1-style API. This mismatch is pre-existing and unrelated to this change (the `hardhat-test` CI job is already commented out in `.github/workflows/backend_integration_tests_pr.yml`). I verified the Solidity change by careful manual review instead — tracing every call site of `_addTransaction` in both contracts to confirm the new check doesn't affect the internal ghost-to-ghost message relay paths (`_processMessages`/`_processInternalMessages`/`addInternalMessageTransaction`), only the externally-reachable `addTransaction` and fee-aware fallback entry points.

# Decisions made

- Placed the check in the external `addTransaction` functions (and the fee-aware fallback) rather than the shared internal `_addTransaction` helper, because the internal message-relay paths legitimately call `_addTransaction` with `_sender` ≠ `msg.sender` (a validator finalizing a transaction, relaying on behalf of a ghost) — putting the check in the internal helper would have broken that.
- Scope note (see [comment below](#issuecomment-5695630350)): CodeRabbit correctly flagged that the `ghostContracts` registry this PR's carve-out relies on is itself writable by any caller today, via the unauthenticated `emitTransactionAccepted`/`emitTransactionFinalized` entry points. That's tracked separately as #1767 and is a larger, separate fix (gating those entry points, validating ghost registration) — intentionally not bundled into this PR, which targets the narrower #1768 sender-spoofing bug.

# Checks

- [x] I have tested this code

# User facing release notes

Fixes a bug where a transaction's sender/caller identity could be spoofed: the actual signer (who pays gas) and the address seen by the Intelligent Contract as the caller could differ, bypassing any contract logic that gates behavior by caller address.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved transaction security by rejecting transactions when the declared sender does not match the recovered signer.
  - Added authorization checks for transactions submitted on behalf of another sender.
  - Empty transaction submissions now fail immediately with a clear error.

- **Tests**
  - Added coverage for valid sender matching and rejection of sender spoofing attempts.


bars26:
Re: the critical merge-risk note above (\`ghostContracts\` registry being writable by any caller) — good catch, but that's a separate, broader issue than what #1768 (and this PR) targets.

\`ghostContracts\` is currently populated via \`_storeGhost\`, reachable from the unauthenticated \`emitTransactionAccepted\` / \`emitTransactionFinalized\` entry points (no access-control modifier on either). An attacker could in principle get their own address registered as a ghost and use that to satisfy this PR's `ghostContracts[msg.sender]` carve-out. That gap is tracked separately as #1767 (unauthenticated internal-message emit helpers, ungated `Transactions` lifecycle, and ungated `ConsensusManager` seed writes) — it needs its own fix (gating those two entry points, validating ghost registration against a trusted source) that's larger and riskier than this PR's scope.

This PR closes the specific bug in #1768: \`_sender\` being trusted from calldata with no check against \`msg.sender\`/the verified signer, in both the externally-reachable \`addTransaction\` entry points (including the fee-aware fallback I found needed the same fix while rebasing) and the Python backend's transaction decoding. I'd rather not bundle #1767's fix into this PR — flagging it here so maintainers can prioritize it separately.

